PR #11, opened 2017-04-14
feat(setup): Top-level setup page

lwander:


setup/quickstart/solutions.md
```diff
@@ -0,0 +1,16 @@
+---
+layout: single
+title:  "Solutions"
+sidebar:
+  nav: setup
+---
+
+There are a few Quickstart solutions:
+
+* [AWS](https://aws.amazon.com/about-aws/whats-new/2016/08/netflix-oss-spinnaker-on-the-aws-cloud-quick-start-reference-deployment/)
```
ttomsu: I'd suggest expanding this out - "Amazon Web Services"

setup/quickstart/solutions.md
```diff
@@ -0,0 +1,16 @@
...
+* [Azure](https://aka.ms/azspinnaker)
```
ttomsu: I'd suggest using the full brand name - "Microsoft Azure". Our brand people would get touchy too if we dropped "Google" from "Google Cloud Platform"

ttomsu (approved):

setup/quickstart/solutions.md
lwander: Good call

setup/quickstart/solutions.md
lwander: Makes sense

PR #13, opened 2017-04-17
Makes sidebar and top nav 'activate' when user is on that page or a s…

ttomsu:
…ubpage.

ohmnka6hrek

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -1,187 +1,187 @@
 main:
   - title: Concepts
-    url: /concepts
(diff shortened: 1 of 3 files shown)


lwander (approved):
Sweet

lwander:
For consistency, what do you think about highlighting the sidebar element (e.g. Getting Started) by bolding it in the same way you do for the top nav?

ttomsu:
Lars and I looked at the difference and decided to stick with the colored background + white text

PR #20, opened 2017-04-19
Update to nav structure for Community section

skim1420:


_data/navigation.yml
```diff
@@ -146,22 +147,32 @@ reference:
 community:
   - title: Releases
     url: /community/releases/
```
lwander: Do we will want top-level clickable URLs for these?

lwander (approved):

skim1420:
Right, updated

PR #44, opened 2017-04-26
Adds halyard getting started codelab

skim1420:
@spinnaker/google FYI

### DIFF
--- a/guides/tutorials/codelabs/halyard-getting-started/index.md
+++ b/guides/tutorials/codelabs/halyard-getting-started/index.md
@@ -21,22 +21,77 @@ In our scenario, we want to create a Spinnaker instance and set it up as follows
 For this exercise we will be operating entirely within one GCP project, and use Google Container Engine (GKE) as our Kubernetes cluster. 
 
 
-## Part 0: Prerequisites
+## Part 0: Preparation
 
-We assume the reader has some prior experience with Kubernetes and GCP, but will also provide some guidance along the way. First, you'll need to [install gcloud](https://cloud.google.com/sdk/downloads#interactive) and have a [GKE](https://console.cloud.google.com/kubernetes/list)) cluster.
+### Install gcloud
 
+If you don't already have gcloud installed, navigate to [Installing Cloud SDK](https://cloud.google.com/sdk/downloads#interactive) to install gcloud
 
-## Part 1: Install halyard
+### Create a Kubernetes cluster
+
+Navigate to the [Google Cloud Console's GKE section](https://console.cloud.google.com/kubernetes/list) to create a new Kubernetes cluster (please note the cluster name and zone).
+
+### Enable the GCP IAM API
+
+Navigate to the [Google Cloud Console](https://console.developers.google.com/apis/api/iam.googleapis.com/overview) and enable the Google Identity and Access Management (IAM) API
+
+### Set up credentials
+
+Create a service account for our halyard host VM:
+
+```
+GCP_PROJECT=$(gcloud info --format='value(config.project)')
+HALYARD_SA=halyard-service-account
+
+gcloud iam service-accounts create $HALYARD_SA \
+    --project=$GCP_PROJECT \
+    --display-name $HALYARD_SA
+
+HALYARD_SA_EMAIL=$(gcloud iam service-accounts list \
+    --project=$GCP_PROJECT \
+    --filter="displayName:$HALYARD_SA" \
+    --format='value(email)')
+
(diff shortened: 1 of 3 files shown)


guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+### ~/.kube/config
+
+If you don't already have this, an easy way to set this up is to use gcloud:
+
+    gcloud container clusters get-credentials {YOUR_GKE_CLUSTER_NAME}
+
+### GCP service account
+
+You will need to specify the location of a json file containing credentials to your GCP project with the role "Storage / Storage Admin". Download the key from the Cloud Console and copy it to `~/.google/account.json`.
```
skim1420: @ewiseblatt This service account will be used for both front50/GCS as well as GCR. Not sure if there's a lesser privilege that will suffice for the needs.

guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -0,0 +1,135 @@
...
+## Part 1: Install halyard
+
+### Create halyard host VM
+
+```
+HALHOST=$USER-halyard-`date +%m%d`
+
+gcloud compute instances create $HALHOST \
+    --scopes storage-full,compute-rw \
+    --project={YOUR_GCP_PROJECT} \
```
lwander: I think it's OK to drop this, or to add a call like this

```
PROJECT=$(gcloud info --format='value(config.project)')
```

to make this guide more copy/paste-able

guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Part 1: Install halyard
+
+### Create halyard host VM
+
+```
+HALHOST=$USER-halyard-`date +%m%d`
+
+gcloud compute instances create $HALHOST \
+    --scopes storage-full,compute-rw \
```
lwander: I don't think any scopes are needed on this VM if you're just running Halyard to deploy remotely.

guides/tutorials/codelabs/halyard-getting-started/index.md
lwander: If the VM has the right scopes, you can download the keys straight to the VM with the commands shown here: https://spinnaker.github.io/setup/providers/docker-registry/#google-container-registry

guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+. ~/.bashrc
+```
+
+## Part 2: Gather needed credentials
+
+### ~/.kube/config
+
+If you don't already have this, an easy way to set this up is to use gcloud:
+
```
lwander: Don't forget to add 
```
gcloud config set container/use_client_certificate true
```
first

guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -0,0 +1,135 @@
...
+    gcloud container clusters get-credentials {YOUR_GKE_CLUSTER_NAME}
+
+### GCP service account
+
+You will need to specify the location of a json file containing credentials to your GCP project with the role "Storage / Storage Admin". Download the key from the Cloud Console and copy it to `~/.google/account.json`.
+
+
+## Part 3: Set Spinnaker configuration
+
+We will install Spinnaker v0.1.0
```
lwander, line 171: (Optimist Pram)

lwander (approved):
Thanks for setting this up - I'm wondering if instead we should move this under "quickstarts" or something in "setup", since that is probably where the user will look for something like this first.

guides/tutorials/codelabs/halyard-getting-started/index.md
ewiseblatt: I'm not following this discussion or concerns.

Aside, where does ~/.google/account.json come from or get used?
I have seen "standard" places for credentials, but that isnt one I recall.
We dont use this, but the Google client libraries actually defines an environment variable GOOGLE_APPLICATION_CREDENTIALS that points to application default credentials.

It looks like you are putting this here so you can register it later. That's fine. The .google directory sounds like a standard and "account.json" sounds pretty generic, but it's purpose is quite specific.

BTW, the file should be made secure
chmod 400 ~/.google/account.json

skim1420:
All comments addressed

PR #45, opened 2017-04-26
feat(setup/storage): GCS storage setup

lwander:


lwander:

PR #47, opened 2017-04-26
feat(nav): Change active style

lwander:
I changed the color to be consistent with the navigation theme that seems to center around using that blue for breadcrumbs: 

# Before
mqw6undskuq

# After
dnfbeb1koqh


lwander:
Let me know what you think @ttomsu

lwander:
approved by ttomsu in person

PR #64, opened 2017-05-01
docs(halyard): Adds small architecture diagram to the Halyard codelab

ttomsu:


ewiseblatt (commented):
I know this is just an example, however would it be "normal" to be deploying into the same cluster as Spinnaker or would you typically keep the cluster under management separate? Is the codelab clear about that so when people copy it as a starting point they are headed down a good path?

lwander (approved):

lwander:
I think this is fine because it's scoped to the codelab, maybe a disclaimer that this is only one possible outcome could help.

guides/tutorials/codelabs/halyard-getting-started/index.md
```diff
@@ -20,6 +20,11 @@ In our scenario, we want to create a Spinnaker instance and set it up as follows
 
 For this exercise we will be operating entirely within one GCP project, and use Google Container Engine (GKE) as our Kubernetes cluster. 
 
+{% include figure
```
skim1420, line 23: Just curious, why not just embed the image via markdown, i.e. `?`

skim1420: Also, does the markup-based diagramming not support stuff like this?

ttomsu: I could - I wanted a caption though, and the `include figure` provides that with less fuss. Take a look at https://github.com/spinnaker/spinnaker.github.io/blob/master/_includes/figure

PR #99, opened 2017-05-12
docs(api): reformat API doc templates

danielpeach:
The generated docs look like this: 
api_docs

(The double slash in the endpoint will be fixed by a Kork update).

There's a lot of room for improvement, but that will come from improving Gate's controller annotations.
Also fixes the breadcrumbs issue for /reference/api 

@spinnaker/google-reviewers


ttomsu:
That looks great.

PR #114, opened 2017-05-24
fix(install): GCE needs serviceAccountActor

lwander:
I missed this role


lwander:
Also @skim1420 if you run that added command against the service account you used to configure GCE, it should work without having to redeploy.

skim1420 (approved):

PR #118, opened 2017-05-24
docs(pipelinetemplate): Adding initial draft of pipeline templates

robzienert:
First draft through, would love some feedback. It's a big topic so I was trying to weigh bringing in too much vs simplicity. Add, remove, edit, whatever. Bigger picture / direction comments also welcome.

The largest thing of note here is that I explicitly call out usage of my app [tiller](https://github.com/robzienert/tiller). I think there could be a larger discussion around this app's future (moving into spinnaker org?), or what have you - not sure if this PR is the place for it (maybe make a spinnaker/spinnaker issue of it?)

@spinnaker/reviewers PTAL

### DIFF
--- a/guides/tutorials/codelabs/managed-pipeline-templates/index.md
+++ b/guides/tutorials/codelabs/managed-pipeline-templates/index.md
@@ -1,0 +1,245 @@
+---
+layout: single
+title:  "Managed Pipeline Templates"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+This codelab provides an introduction to the Managed Pipeline Templates feature
+of Spinnaker. We'll be creating a basic template that can be used to bake an AMI
+and deploy to AWS and explore how to use the template within Spinnaker.
+
+# Prerequisites
+
+- Have [Pipeline Templates enabled][enable-pipeline-templates] in Orca
+- (Optional) Familiarize yourself with the [Pipeline Templates Spec][spec]
+- Download the latest release of [Tiller][tiller-repo]
+
+If you have not already enabled pipeline templates, you can do so by enabling it
+in `orca.yml`:
+
+```yaml
+pipelineTemplate:
+  enabled: true
+```
+
+# Writing the Template
+
+First, here's the template, we'll break it down below.
+
+{% raw %}
+```yaml
+schema: "1"
+id: defaultDeploy                                                            (1)
+protect: false
+metadata:                                                                    (2)
+  name: Default Deploy
+  description: The default deployment pipeline for Acme Corp applications.
+  owner: example@example.com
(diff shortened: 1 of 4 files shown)


guides/tutorials/codelabs/managed-pipeline-templates/index.md
```diff
@@ -0,0 +1,245 @@
...
+Template Loaders:
+
+* `file`: Load a pipeline template off the machine running Orca.
+* `http`: Load a pipeline template from an unauthenticated HTTP resource.
+* `spinnaker`: Load a pipeline template from Spinnaker itself.
+
+In this section, we'll focus on the recommended loader, `spinnaker`, as it
+provides the greatest gamut of features and flexibility. To publish into 
+Spinnaker, we'll use [Roer][roer-repo], a CLI app for Spinnaker that has
+commands for interacting with Pipeline Templates.
```
lwander, line 176: Why are we creating another CLI?

robzienert: The goal of roer is strictly to give a thin wrapper around the API to make getting pipeline templates and so-on into Spinnaker easier, whereas Halyard's scope is dramatically larger in that it's focused on configuration & operations of Spinnaker itself. If someone is just looking to have a client to publish templates from a CI job, having a zero-dependency tool has a lot of advantages in my mind.

lwander: Can you elaborate on zero-dependency?

A lot of what you'll need to support for this to work (talking to an authenticated spinnaker, submitting and tailing pipeline executions, knowing where gate is installed) is already supported in Halyard. 

Once the KMS stuff is complete in Halyard you'll be able to run

```bash
sudo bash InstallHalyard.sh

hal backup restore # fetches encrypted config

hal pipeline submit ...
```

from any environment, which feels pretty straight-forward/low-dependency to me.

I'm concerned with duplicating work, as well as forcing the community to install a number of CLIs to get Spinnaker to work.

robzienert: Zero dependency, as far as the only thing that needs to be installed is a single binary. Halyard requires the JVM and further yet, requires Ubuntu if they're using `InstallHalyard.sh` (which could be refactored to support other environments). I'm not a fan of a proliferation of CLIs either, but the differences in what roer is attempting to do is decidedly less than halyard and is specifically intended to be lightweight.

As far as duplication of work goes, that's a very real concern, and one I was attempting to address by limiting the scope of roer to strictly DCD things. This CLI was initially being built just for Netflix, but felt it provided enough baseline value for where pipeline templates are to be provided for the entire community.

PR #131, opened 2017-05-30
docs(setup): old aws setup content

skim1420:
Submitting current/old AWS setup content here while we work on halyard support


lwander:
https://github.com/spinnaker/spinnaker.github.io/pull/153

PR #150, opened 2017-05-31
Link to related codelab from Kubernetes page

rcarun:


setup/providers/kubernetes.md
```diff
@@ -36,7 +36,7 @@ hosted solutions:
 
 * [Google Container Engine](https://cloud.google.com/container-engine/)
 * [Azure Container
-  Service](https://docs.microsoft.com/en-us/azure/container-service/container-service-kubernetes-walkthrough)
+  Service](https://docs.microsoft.com/en-us/azure/container-service/container-service-kubernetes-walkthrough) [(codelab)](http://aka.ms/azspinlab)
```
ejizba, line 39: I don't think linking to the codelab here makes sense. This page assumes you have an existing Kubernetes cluster and you are setting up Spinnaker separately. The sentence preceding this even implies you just want to create a cluster by itself

That being said, the existing link is a pretty heavy doc page and we might be able to find a better link

lwander:
This feels out-of-place since the links here are for setting up k8s alone in the context of getting spinnaker up and running. Injecting a codelab that does most of the basic spinnaker config for you without Halyard in the middle of it doesn't seem correct

setup/providers/kubernetes.md
lwander: jinx

ttomsu:
What's the word on this PR?

lwander:
Closing for now because the PR is stale and link seems to be in the wrong place. Feel free to re-open or just create another PR if you want to link to the codelab elsewhere.

PR #156, opened 2017-06-01
docs(gce): Added small section for lb metadata.

jtk54:


reference/providers/gce.md
```diff
@@ -53,14 +53,19 @@ Deploys a GCE managed instance group.
 A new GCE [instance template](https://cloud.google.com/compute/docs/instance-templates) is created for each new managed
 instance group.
 
+If a deployed server group is load-balanced, Spinnaker uses [instance metadata](https://cloud.google.com/compute/docs/storing-retrieving-metadata)
+to store the relationship between the server group and load balancer. This is necessary to represent the object models in Spinnaker's UI.
```
duftler: Not just for proper rendering, right? Also so we can 'remember' the associations even when the server groups are not 'in service'.

duftler (approved):
LGTM after addressing comment.

PR #207, opened 2017-06-08
Fix Links to Codelabs

madisona:
The links to the codelabs are currently broken. Without the trailing slash the links are being redirected to just https://www.spinnaker.io/


ttomsu:
Thanks for the PR. This was actually a problem with our CDN config - it should be fixed now. That being said, adding the slash actually removes a redirect, so I'm going to merge this with much appreciation :-)

PR #211, opened 2017-06-08
feat(docs): Adds redirect-from plugin to map old URLs to their new co…

ttomsu:
…unterparts (if applicable).

@ajordens FYI


ajordens:
Ahh neato.

Just have to manually map the old slugs to new I guess.

PR #223, opened 2017-06-12
feat(versions): Explain top-level versions

lwander:
@spinnaker/google-reviewers

### DIFF
--- a/community/releases/versions.md
+++ b/community/releases/versions.md
@@ -5,6 +5,22 @@ sidebar:
   nav: community
 ---
 
+The Spinnaker releases listed below are top-level versions tying together each
+Spinnaker subcomponents' versions. These have been validated together in an 
+end-to-end integration test suite curated by the Spinnaker community, and
+represent a more maintainable, easy to upgrade Spinnaker. While 
+it's still possible to install each Spinnaker subcomponent's versions 
+independently, there is no guarantee that they will work together.
+
+If you want to see what each top-level version is comprised of, run 
+
+```bash
+hal version bom <version>
+```
+
+to see the commit hash and tag matching `version-<version>` of each
+subcomponent.
+
 ## Latest Stable
 {% assign reversed = site.changelogs | reverse |  %}
 {% for post in reversed %}


skim1420 (approved):

community/releases/versions.md
```diff
@@ -5,6 +5,22 @@ sidebar:
   nav: community
 ---
 
+The Spinnaker releases listed below are top-level versions tying together each
+Spinnaker subcomponents' versions. These have been validated together in an 
+end-to-end integration test suite curated by the Spinnaker community, and
+represent a more maintainable, easy to upgrade Spinnaker. While 
+it's still possible to install each Spinnaker subcomponent's versions
```
ttomsu, line 12: Link for those wanting to learn how to modify the component versions?

ttomsu (approved):

community/releases/versions.md
lwander: What do you mean? Publish your own BOM?

ttomsu: Right - is there documentation on that? If so, I'd link it here.

lwander: There is not - and I've only seen it asked about once by a user with very unique constraints. It's kind of an advanced topic and this doesn't feel like the right place to bring it up.

lwander: We should (at some point) publish comprehensive release docs that address this.

ttomsu: I would then remove the mention that it's possible, and let users ask for it when appropriate.

lwander:
https://github.com/spinnaker/spinnaker/issues/1742

community/releases/versions.md
lwander: Sorry merged before I saw your comment. 

The mention alludes to direct downloads (which users are currently doing).

lwander: The point is for this to contrast the existing install path (downloading the packages via whatever path users are going by) against the Halyard install path. It's not explaining that Halyard can allow you to download whatever versions you want.

PR #242, opened 2017-06-18
Add missing command to AWS provider documentation

serialx:
You need to `hal config provider aws enable` in order to enable AWS support after setting up the account.


lwander (approved):
Thanks :)

PR #264, opened 2017-07-01
docs(fix): change reference to spinnaker/reviewers

tomaslin:
spinnaker/reviewers is not a public group so people without admin access won't be able to reference this


ttomsu:
LGTM

PR #268, opened 2017-07-07
Adding port mapping table to Architecture page

edwinavalos:
Adding a simple table with port information that use to be on the old page. IIRC there was a third column, but I do not remember it's content.

My editor seems to be very opinionated on line endings let me know if you need me to revert the changes on line 16 and 69-70


medyagh:
+1

lwander (approved):
Thanks

PR #270, opened 2017-07-11
Update command for enabling Fiat

msloes:
Was: `hal config security roles enable` which fails/outputs: `Expected a command, got roles`

Correct command: `hal config security authz enable`


ttomsu:
:+1: Thank you!

PR #279, opened 2017-07-25
docs(setup) fix redirected notification links

erran:
# Context
This uses relative links and properly links to the section anchors.

# Before
spinnaker-bad-links

# After
spinnaker-fixed-links


aglover:
Thank you! LGTM.

PR #282, opened 2017-07-26
feat(halyard): List cluster resource reqs

lwander:
This is a rough guess based on what we recommend for the single-vm install in cloud launcher.


aglover:
LGTM

PR #293, opened 2017-08-02
halyard.md: Improve to the Docker run command

cristim:
Changes:
- run container with --rm 
- give the container a name and use it in the subsequent `docker run` command.
- hint additional volume mounts for usage with kubernetes.


lwander:
Very nice, thank you :)

lwander (approved):

PR #340, opened 2017-09-08
fix(changelog): Include note on Halyard version requirement for v1.3.0

ttomsu:


jtk54 (approved):
LGTM. You can also manually add the chore section to that gist to explain the spring upgrades.

PR #349, opened 2017-09-18
docs(setup): Show how to setup authn x509 roleOid parsing.

jasondtgt:


edwinavalos (approved):
Approved but will wait till the feature is available in halyard and gate

PR #355, opened 2017-09-22
Remove localhost prepend from docker run command

mononokehime:
When specifying the ports to expose, there is no need to prepend localhost. The command doesn't run in docker 17.06. Here is the output:
docker run -p localhost:8084:8084 -p localhost:9000:9000 \
>     --name halyard --rm \
>     -v ~/.hal:/root/.hal \
>     gcr.io/spinnaker-marketplace/halyard:stable
docker: Invalid ip address: localhost.
See 'docker run --help'.

### DIFF
--- a/setup/install/halyard.md
+++ b/setup/install/halyard.md
@@ -70,7 +70,7 @@ Now, run the Halyard docker container, while mounting that Halyard config
 directory for your container:
 
 ```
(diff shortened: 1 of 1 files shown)


lwander:
Closing as this section has been rewritten.

PR #359, opened 2017-09-23
Add DC/OS as a cloud provider in navigation

cchesser:
Adding the DC/OS cloud provider to the navigation bar. Support for this cloud provider was added in [1.3.0 release](https://www.spinnaker.io/community/releases/versions/1-3-0-changelog).


duftler (approved):
LGTM.

PR #386, opened 2017-10-26
Add South Bay Meetup

benkebhb:
Hi - here are South Bay Meetup changes from bbenke@cisco.com. Note the videos are on a Cisco public site.  Its not the same pattern you have now where they are on one you tube site.   I do have the original source .mpg files if someone needs for a you tube site but they are huge files so not sure how easy to transfer.


aglover:
LGTM

PR #391, opened 2017-10-31
Add link to other providers for clarity

nneul:


dorbin:
@nneul  It looks like you've added a link to the superset of providers, but only the two already listed support distributed installation, so it would be misleading to add this link. 

Let me know if I'm missing something about your intent :)

nneul:
Ah, I'm just getting up to speed and was thrown off by them not being mentioned. If indeed they don't support that portion of the config, my PR doesn't make sense.

PR #418, opened 2017-11-21
docs(K8s): stubs for Kubernetes v2 docs

dorbin:
Just some stubs for you to get started with, Lars


lwander (commented):
Can you change the file names to treat "kubernetes" consistently? I'm not sure what the existing convention is, but these seem to be mixed between `k8s` and `kub`

dorbin:
Yes, but there is already "kub" and "kubernetes" before this change. Should I just change me "k8s"s to "kub"s and let the existing inconsistencies stand?

lwander:
Ah shame, let's leave the existing stuff as-is and replace k8s with kub

reference/providers/k8sv2/index.md
```diff
@@ -0,0 +1,12 @@
+---
+ Layout: single
```
lwander: This directory seems to follow the k8s naming rather than kub

lwander (approved):
Thanks!

lwander:
(Just waiting for the build)

Issue #413, opened 2017-11-17
Add a "minio" section under "persistent storage"


lwander:
Since the redis storage source is so unstable, we need another local option with installation steps available.

PR #420, opened 2017-11-21
feat(minio): demonstrate minio configuration

lwander:
Closes https://github.com/spinnaker/spinnaker.github.io/issues/413

### DIFF
--- a/setup/storage/minio.md
+++ b/setup/storage/minio.md
@@ -1,0 +1,41 @@
+---
+layout: single
+title:  "Minio"
+sidebar:
+  nav: setup
+---
+
+> :warning: Losing Minio's data will mean losing all your Spinnaker 
+> application metadata, and configured pipelines.
+
+[Minio](https://www.minio.io/) is an S3-compatible object store that you can
+host yourself. This is the persistent storage solution we recommend when you
+don't want to depend on a cloud provider to host your Spinnaker data. 
+
+## Prerequisites
+
+Install Minio following the instructions on the [Minio
+homepage](https://www.minio.io/), making sure to have it run on an endpoint
+reachable by Spinnaker. Record the following values:
+
+* `ENDPOINT`: The fully-qualifed endpoint Minio is reachable on. If Minio is
+  running on the same machine as Spinnaker, this might be
+  `http://localhost:9001`.
+
+* `MINIO_ACCESS_KEY`, `MINIO_SECRET_KEY`: The access/secret keypair you've 
+  configured Minio with. These env vars need to be visible to the Minio process
+  for them to work.
+
+## Editing Your Storage Settings
+
+Run the following commands (notice we are picking S3 as our storage type,
+because Minio implements the S3 API):
+
+```bash
+echo $MINIO_SECRET_KEY | hal config storage s3 --endpoint $ENDPOINT \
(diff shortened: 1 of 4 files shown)


setup/storage/redis.md
```diff
@@ -5,6 +5,11 @@ sidebar:
   nav: setup
 ---
 
+> :warning: We _highly_ recommend relying on [Minio](/setup/storage/minio)
+> instead of Redis if you are looking for a local persistant storage solution
+> for Spinnaker. The Redis storage implementation is untested and unsupported 
+> by anyone in the Spinnaker community right now.
```
dorbin: Delete "right now"

setup/storage/minio.md
```diff
@@ -0,0 +1,40 @@
...
+layout: single
+title:  "Minio"
+sidebar:
+  nav: setup
+---
+
+[Minio](https://www.minio.io/) is an S3-compatible object store that you can
+host yourself. This is the persistent storage solution we recommend when you
+don't want to depend on a cloud provider to host your Spinnaker data. However,
+losing Minio's data will mean losing all your Spinnaker application metadata,
```
dorbin: Maybe break out this last sentence, "However, losing ....pipelines" into a warning paragraph.

setup/storage/redis.md
lwander: SG

setup/storage/minio.md
```diff
@@ -0,0 +1,40 @@
...
+  `http://localhost:9001`.
+
+* `MINIO_ACCESS_KEY`, `MINIO_SECRET_KEY`: The access/secret keypair you've 
+  configured Minio with. These env vars need to be visible to the Minio process
+  for them to work.
+
+## Editing Your Storage Settings
+
+Run the following commands (notice we are picking S3 as our storage type,
+because Minio implements the S3 API):
```
dorbin, line 32: Does Minio support other storage types? If not, do we need this "(notice we are...)" ? And if so, should we mention somewhere near the beginning the different storage types it can be used with?

lwander: My worry is that this looks like a bug/documentation error with `--type s3`. It's not clear why that works, and pointing it out right above the command feels like it might avoid some confusion.

setup/storage/minio.md
lwander: Good point

setup/storage/minio.md
lwander: Also, it only implements the S3 API AFAIK

dorbin: According to the minio.io/ they're "multi cloud," though they also mention the s3 api (and *only* the s3 api)  right at the top of the page. 
"Minio is an open source object storage server with Amazon S3 compatible API. 
Build cloud-native applications portable across all major public and private clouds"
Do we anticipate anyone using minion who's not using s3? 

I agree, btw, that if you have the `type s3` option and don't point it out, it looks like a mistake.

lwander: Minio replaces S3, so it's not about which cloud they are using. For example, I know there are subtle difference between S3 and GCS, so while Minio could (in theory) replace GCS, I don't think it's smart to try.

PR #424, opened 2017-11-27
fix(pubsub): Add example message and more explanation on templates.

jtk54:


guides/tutorials/codelabs/pubsub-to-appengine/index.md
```diff
@@ -97,13 +97,41 @@ Now configure Spinnaker to receive messages from your Google Cloud Pub/Sub subsc
...
+      "etag": "CIu+09aj39cCEAE="
+    }
+    ```
+
+    Any of the keys present in this message can be used in the transformation definition. Jinja is expressive, so you can even include
+    things like loops, array indices, and paths to sub-objects in your template. The only constraint on the template file is that
+    the resultant object must match [the artifact model class](https://github.com/spinnaker/kork/blob/master/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java),
+    and that one must be supplied to Spinnaker.
+
     Store the path to 'gcs-jinja.json' as an environment variable for the next step:
```
dorbin: With the temlate and the sample payload json, plus a few sentences after, there's a lot of stuff between step 2 and step 3.  Reader might get a little lost. One small improvement might be to make the "Store the path to..." and "Make sure the file permissions..." are substeps, a. and b. That will at least highlight them as things to do, and descrease the risk they'll get missed as just a couple more paragraphs of explanation.

dorbin (approved):
LGTM

guides/tutorials/codelabs/pubsub-to-appengine/index.md
jtk54: Clarified that section with substeps.

PR #440, opened 2017-12-07
fix(local git): mention required node version

lwander:


brandonnelson3 (approved):

setup/install/environment.md
```diff
@@ -87,6 +87,7 @@ Ensure that the following are installed on your system:
 * git
 * curl
 * yarn ([guide](https://yarnpkg.com/lang/en/docs/install/))
+* node (version >=8.9.0, [can be installed via nvm](https://github.com/creationix/nvm#install-script))
```
ttomsu, line 90: You should install node & npm first, which you can then install yarn using `npm` (I had to do this yesterday)

lwander: Ah nice, sounds good. Can you submit a PR with the steps you took yesterday?

PR #451, opened 2017-12-13
Update s3.md with reference to minio

dreynaud:
Minio is only mentioned on the redis setup page, it may make sense to also mention it on the S3 page since it is S3-like


lwander (approved):

lwander:
SGTM

PR #457, opened 2017-12-15
Correct For New DIV in Jekyll Syntax Markup

ghost:
Jekyll 3.6.2 appears to introduce an extra div tag wrapping syntax highlight blocks.  Locally testing with Jekyll 3.4.4 does not reproduce the issue but updating to 3.6.2 demonstrates it.  It's also visible on the production github pages site.

This new markup results in syntax blocks that can appear overly tall when aligned with the base of the "On This Page" navigation section.

This commit introduces a CSS workaround that fixes the visual issue but does not address the extra markup being put into the rendered HTML.

### Before

before_extra_height_fix

### After

after_extra_height_fix

### DIFF
--- a/_sass/_syntax.scss
+++ b/_sass/_syntax.scss
@@ -31,6 +31,11 @@ figure.highlight {
     margin: 0;
     padding: 1em;
   }
+
+  div.highlight {
+    padding: 0;
+    display: inline-block;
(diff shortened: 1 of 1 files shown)


lwander (approved):
Nice!

PR #458, opened 2017-12-15
Add missing authz config for github groups

bfarrell:
Missing step which leaves service as EXTERNAL.  Need to set it as github.


lwander (approved):

lwander:
Thanks!

PR #461, opened 2017-12-18
feat(provider/kuberentes): v2 workload documentation

lwander:


reference/providers/kubernetes-v2/index.md
```diff
@@ -114,8 +114,117 @@ Kubernetes supports. Also the Kubernetes extension mechanisms&mdash;called
...
+It is worth noting that the resource mapping exists primarily to render
+resources in the UI according to Spinnaker conventions, it does not affect how
+resources are deployed or managed. Instead, this is determined by the following
+properties:
+
+* __Versioning__
+
+  If a resource is "versioned", it will always be deployed with a new sequence
+  number `vNNN`. This is important for resources like ConfigMaps and
+  ReplicaSets, which don't have their own builtin update policy like
```
duftler: s/builtin/built-in

reference/providers/kubernetes-v2/index.md
```diff
@@ -114,8 +114,117 @@ Kubernetes supports. Also the Kubernetes extension mechanisms&mdash;called
...
+  * _Undo Rollout:_ (Workloads only) Can this resource be rolled back/forward
+    to an existing revision?
+  * _Pause Rollout:_ (Workloads only) When rolling out, can the rollout be
+    stopped?
+  * _Resume Rollout:_ (Workloads only) When a rollout is paused, can it be
+    started again?
+
+* __Stability__
+
+  This describes under what conditions is this kind considered stable after a
```
duftler: s/is this kind considered stable/this kind is considered stable

reference/providers/kubernetes-v2/index.md
```diff
@@ -114,8 +114,117 @@ Kubernetes supports. Also the Kubernetes extension mechanisms&mdash;called
...
+  new `spec` has been submitted.
+
 ## Workloads ≈ Server Groups
 
+Anything classified as a Spinnaker Server Group will be rendered on the
```
duftler: Feels like the information below would be easier to consume if presented via a table.

duftler (approved):
LGTM after addressing comments.

reference/providers/kubernetes-v2/index.md
lwander: Done

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
-# Resource Mapping
+# How Kubernetes Resources Are Managed by Spinnaker
 
-Resource mapping between Spinnaker and Kubernetes constructs is a lot more
-flexible than for other providers, because of how many types of resources
-Kubernetes supports. Also the Kubernetes extension mechanisms&mdash;called
-[Custom Resource Definitions
+Resource mapping between Spinnaker and Kubernetes constructs, as well as the
+introduction of new types of resources, is a lot more flexible in the
+Kubernetes provider V2 for other providers, because of how many types of
```
dorbin, line 111: "*than* for other providers"

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+resources Kubernetes supports. Also the Kubernetes extension
```
dorbin, line 112: "...because v2 takes advantage of the many resource types Kubernetes supports" 

Because k8s's support for many resource types was already in play, right, but v1 didn't exploit that? IOW, you're crediting K8s with the support, but you're comparing v2 with other providers, including k8sv1.

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+mechanisms&mdash;called [Custom Resource Definitions
 (CRDs)](https://kubernetes.io/docs/concepts/api-extension/custom-resources/)&mdash;make
 it easy to build new types of resources, and Spinnaker accomodates that by
 making it simple to extend Spinnaker to support a user's CRDs.
 
-## Workloads ≈ Server Groups
+## Terminology Mapping
 
-## Services, Ingresses ≈ Load Balancers
+It is worth noting that the resource mapping exists primarily to render
```
dorbin, line 120: I think you can delete "It is worth noting that the"

But also, what do you mean by "primarily"?

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+resources in the UI according to Spinnaker conventions. It does not affect how
+resources are deployed or managed.
 
-## NetworkPolicies ≈ Security Groups
+There are three major groupings of resources in Spinnaker, Server Groups, Load
```
dorbin, line 124: s/,/:

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+Balancers, and Security Groups. They correspond to Kubernetes resource kinds as
+follows:
+
+* Workloads ≈ Spinnaker Server Groups
+* Services, Ingresses ≈ Load Balancers
```
dorbin, line 129: What is the difference between services and ingresses here? I don't know enough about them to accept that they _both_ map to load balancers. But the distinction might not be important here; I leave it to you.

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+* NetworkPolicies ≈ Security Groups
```
dorbin, line 130: In this list, should you swap positions? Because you first list the groupings of resources on the Spinnaker side, above, it's confusing that the K8s resources are on the left of the mappings here.

Also, I don't think all the capitalizations are needed here.

reference/providers/kubernetes-v2/index.md
```diff
@@ -104,19 +104,85 @@ Serveral annotations are built into each manifest and cannot be used otherwise.
...
+  redeploying can have unexpected results, and delete prior history.
+  Regardless, whatever the policy is, it can be overriden during a deploy
+  manifest stage.
+
+* __Stability__
+
+  This describes under what conditions this kind is considered stable after a
+  new `spec` has been submitted.
+
+## Workloads
```
dorbin, line 171: As we discussed, demote these three h2 headings to h3, and add a wrapper heading to link to from above.

dorbin (commented):
This, plus I'll do an editing run on it after you merge.

reference/providers/kubernetes-v2/index.md
lwander: I'm crediting k8s for the resource types, but pointing out that v2 is the first provider to take advantage of so many (including v1, which could have taken advantage of that many).

reference/providers/kubernetes-v2/index.md
lwander: API endpoints for rendering the resources need to follow this mapping as well - is it worth pointing that out, or should I drop 'primarily'?

reference/providers/kubernetes-v2/index.md
lwander: I think it's fair to leave as-is here. Ingresses just make services reachable outside a cluster. Both load balance traffic to your application.

reference/providers/kubernetes-v2/index.md
lwander: Will swap, makes sense.

reference/providers/kubernetes-v2/index.md
dorbin: Both. Drop "primarily" and point out the api endpoints requirement too.

reference/providers/kubernetes-v2/index.md
dorbin: "...is a lot more flexible in the...V2 than for other providers because it supports more of the many resource types available in Kub..." ?

reference/providers/kubernetes-v2/index.md
dorbin: SG. I do think these can be lower-case tho, right?

PR #462, opened 2017-12-18
Issue 2177

dorbin:


reference/artifacts/in-pipelines/index.md
```diff
@@ -7,28 +7,28 @@ sidebar:
...
-explain how this works, we need to understand the concept of an "expected
-artifact".
+Now that you have an idea of [what an artifact is](../#about-spinnaker-artifacts) in Spinnaker, you need to
+understand how it's used within pipelines. An artifact arrives in a pipeline execution either from an external trigger (for example, code check-in) or by getting fetched by a stage. That artifact is then consumed by downstream stages based on pre-defined behavior. 
+
+Spinnaker uses an "expected artifact" to enable a stage to fetch the needed artifact.
 
 # Expected Artifacts
 
+An "expected artifact" is a specification of what properties (found in the URI decoration) against which to match when searching for the desired artifact.
```
lwander: I'm not sure about this since it also defines fallback behavior

reference/artifacts/in-pipelines/index.md
```diff
@@ -7,28 +7,28 @@ sidebar:
...
+> unexpected behavior.
 
-Once you have an idea of what an "artifact" is in Spinnaker, it's important to
-understand how they are used within pipelines. The idea is simple, artifacts
-arrive in a pipeline execution either from an external trigger or a stage, and
-are consumed by downstream stages based on pre-defined behavior. In order to
-explain how this works, we need to understand the concept of an "expected
-artifact".
+Now that you have an idea of [what an artifact is](../#about-spinnaker-artifacts) in Spinnaker, you need to
+understand how it's used within pipelines. An artifact arrives in a pipeline execution either from an external trigger (for example, code check-in) or by getting fetched by a stage. That artifact is then consumed by downstream stages based on pre-defined behavior.
```
lwander: I would use a different example than "code check-in" since while it's supported, it's a little outside of what Spinnaker specializes in. I would say instead "a docker image pushed to a registry".

lwander (approved):

reference/artifacts/in-pipelines/index.md
```diff
@@ -7,28 +7,28 @@ sidebar:
...
+> history in Redis. Deleting recent executions from Redis can cause
+> unexpected behavior.
 
-Once you have an idea of what an "artifact" is in Spinnaker, it's important to
-understand how they are used within pipelines. The idea is simple, artifacts
-arrive in a pipeline execution either from an external trigger or a stage, and
-are consumed by downstream stages based on pre-defined behavior. In order to
-explain how this works, we need to understand the concept of an "expected
-artifact".
+Now that you have an idea of [what an artifact is](../#about-spinnaker-artifacts) in Spinnaker, you need to
```
lwander: I think this ../ style link is breaking the build. You may need /reference/artifacts/#about-spinnaker-artifacts instead

reference/artifacts/in-pipelines/index.md
```diff
@@ -11,7 +11,7 @@ sidebar:
 > history in Redis. Deleting recent executions from Redis can cause
 > unexpected behavior.
 
-Now that you have an idea of [what an artifact is](/reference/artifacts#about-spinnaker-artifacts) in Spinnaker, you need to
+Now that you have an idea of [what an artifact is](https://www.spinnaker.io/reference/artifacts/#about-spinnaker-artifacts) in Spinnaker, you need to
```
lwander: I don't think this is the right approach, since this breaks if we ever move the hostname.

PR #468, opened 2017-12-22
feat(docs): introduce collapsible nav

ghost:
Allow nav sections in the sidebar to be collapsed by clicking on section titles.

Move "Cloud Providers" top-level nav to be submenu of Install and Configure > Cloud Providers.

cloud-providers-subnav

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -39,7 +39,28 @@ setup:
       - title: Storage
         url: /setup/install/storage/
       - title: Cloud Providers
-        url: /setup/install/providers/
+        collapsed: true
+        children:
+          - title: Overview
+            url: /setup/providers/
+          - title: Amazon Web Services
+            url: /setup/providers/aws/
+          - title: Azure
+            url: /setup/providers/azure/
+          - title: DC/OS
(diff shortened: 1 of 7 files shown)


_data/navigation.yml
```diff
@@ -39,7 +39,28 @@ setup:
       - title: Storage
         url: /setup/install/storage/
       - title: Cloud Providers
-        url: /setup/install/providers/
```
dorbin, line 42: So the Cloud Providers node is no longer hyperlinked to setup/install/providers/index.html right? Is that necessary in order to make it collapse/expandable, or can the expand/collapse be done using an arrow icon only, such that I could click into the Overview from this heading __or__ expand it to see the sub-headings?

ghost: This is technically feasible I think but may prove a bit surprising and fiddly.  The click target for expand / collapse would be small, particularly for tablet users, and the behaviour of a chevron working differently to its label deviates from other examples I can find (e.g. side nav on https://kubernetes.io/docs/home/ and https://cloud.google.com/pubsub/docs/publisher).  Open to discussion here, let me know if you have more thoughts on this.

PR #473, opened 2017-12-28
docs(reference): changed account in description to account name

archana-s:
Updated all references to account in the API to clearly indicate its account name.

Some of our users were confused whether the documentation referred to account name or account id. This can clear up that confusion.


archana-s:
@skim1420 @dorbin Can you please take a look?

skim1420 (approved):

PR #491, opened 2018-01-09
Update docs on kube v2 provider

davidxia:
with necessary commands to update the GUI.

### DIFF
--- a/setup/providers/kubernetes-v2/index.md
+++ b/setup/providers/kubernetes-v2/index.md
@@ -93,6 +93,12 @@ hal config provider kubernetes account add my-k8s-v2-account \
     --context $(kubectl config current-context)
 ```
 
+You'll also need to run
+
+```bash
+hal config features edit --artifacts true
+```
+
 ## Advanced Account Settings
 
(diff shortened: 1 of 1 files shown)


davidxia:
@lwander

setup/providers/kubernetes-v2/index.md
```diff
@@ -93,6 +93,13 @@ hal config provider kubernetes account add my-k8s-v2-account \
     --context $(kubectl config current-context)
 ```
 
+You'll also need to run the following
+
+```
+hal config features edit --artifacts true
+hal deploy apply
```
davidxia: Running `hal deploy apply --service-names deck` and `hal deploy apply --service-names clouddriver front50` didn't work. Redeploying everything did though.

lwander: I would omit the `hal deploy apply` here, since the other docs assume this is part of a larger flow where `hal deploy apply` gets called at the end.

davidxia: Sounds good.

lwander (approved):
Thanks!

PR #493, opened 2018-01-10
Change SELinux context

fabiogoma:
Inside the container /root/.hal directory is a mounted volume, the context is the same from the host, without changing the context inside the container you'll get permission denied messages even with correct user, group and rwx permissions.

### DIFF
--- a/setup/storage/minio.md
+++ b/setup/storage/minio.md
@@ -32,6 +32,7 @@ Run the following commands (notice we are picking S3 as our storage type,
 because Minio implements the S3 API):
 
 # Ubuntu instalation
+
 ```bash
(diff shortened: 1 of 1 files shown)


setup/storage/minio.md
```diff
@@ -39,3 +40,16 @@ echo $MINIO_SECRET_KEY | hal config storage s3 edit --endpoint $ENDPOINT \
 
 hal config storage edit --type s3 
 ```
+
+# Docker container instalation
+```bash
```
lwander: Could you add a new line after these headers? 

Also, please give a callout to the fact that this should be run inside the docker container only.

lwander (approved):
Thanks!

setup/storage/minio.md
lwander: Thanks!

PR #497, opened 2018-01-10
Docs(setup) Big re-org of Setup nav, along with minor re-writes

dorbin:


skim1420:
Sorry for the delay, let's please review together. Hard to get a feel for how well the nav structure works looking at underlying files.

Also, if you could add some set of screenshot, description/jist, rationale summary, etc. that would help too.

dorbin:
@skim1420 ,

The main change here is to the nav for /setup/. Currently it's not procedural at the high level, and it references sections (like bakery and CI) that are lower down on the nav. In fact, the nav is too long, for this reason.

In an earlier PR I moved Storage and Providers up into the Install nav and removed the related sections that were lower down. This PR does something similar to bakery, ci, monitoring, features, and it makes the main install/config procedural. That is, the parts that are stepwise are now numbered. 

After this PR, it'll be divided up into quickstarts, install, and "configure everything else." Additionally, it takes advantage of change Scott made to allow collapsable nav headings. 

Screenshots to come...

dorbin:
setup_nav_collapsed
setup_nav_everything_else
setup_nav_install

skim1420 (approved):

PR #498, opened 2018-01-11
docs(Guide): Update kube v2 minimal RBAC role

davidxia:


setup/providers/kubernetes-v2/index.md
```diff
@@ -59,19 +59,13 @@ metadata:
  name: spinnaker-role
 rules:
 - apiGroups: [""]
-  resources: ["namespaces", "configmaps", "events", "replicationcontrollers", "serviceaccounts", "pods/logs"]
```
davidxia, line 62: Events, replicationcontrollers, serviceaccounts, and pods/logs only seem to be used by v1.

setup/providers/kubernetes-v2/index.md
```diff
@@ -59,19 +59,13 @@ metadata:
...
-  verbs: ["get", "list"]
-- apiGroups: [""]
-  resources: ["pods", "services", "secrets"]
+  resources: ["configmaps", "namespaces", "pods", "secrets", "services"]
```
davidxia: v2 seems to need namespace + delete [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/deployer/KubernetesNamespaceHandler.java#L29).

davidxia:
@ethanfrogers @lwander 

I took the union of permissions v2 needs by looking at code [here](https://github.com/spinnaker/clouddriver/tree/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent) and [here](https://github.com/spinnaker/clouddriver/tree/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/deployer). I removed anything that looked like only v1 needed. I ordered nouns within each apiGroup alphabetically.

I'm not sure about the [`*ManifestOperation`](https://github.com/spinnaker/clouddriver/tree/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest) classes though. Do these classes allow any arbitrary kube noun + verb combination? If so, does that mean that the RBAC actually needs to grant every single permission to Spinnaker?

davidxia:
@ethanfrogers @lwander

setup/providers/kubernetes-v2/index.md
ethanfrogers: I'm pretty sure `pods/logs` is used by V2 as well: 

https://github.com/spinnaker/clouddriver/blob/1376edebf7f6776202e6ec15925e0ea280274b65/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java#L121-L133

ethanfrogers:
other the `pods/logs` thing, LGTM!

setup/providers/kubernetes-v2/index.md
davidxia: Thanks. Added.

lwander (approved):

lwander:
You guys are great, thanks

davidxia:
> I'm not sure about the *ManifestOperation classes though. Do these classes allow any arbitrary kube noun + verb combination? If so, does that mean that the RBAC actually needs to grant every single permission to Spinnaker?

Just curious about this.

ethanfrogers:
@davidxia that's a great point. by my understanding, that is the case.

lwander:
Not quite, things like "ControllerRevisions", or "pods/logs" won't be modified (only read by Spinnaker). A user can also always reduce the RBAC as well depending on what they want any particular Spinnaker account to be able to do.

davidxia:
I see. Since the docs say these are the minimal RBAC config, I think we are
OK here.
-- 
David Xia
davidxia.com

PR #502, opened 2018-01-16
docs(halyard_jars): Add docs for how to install halyard via jars.

brandonnelson3:


setup/install/halyard_migration.md
```diff
@@ -0,0 +1,28 @@
+---
```
dorbin, line 1: I think you'll need to either add a sentence in halyard.md that refers to this doc or add this to _data/navigation.yml. Or  both? Unless you only want this page accessible by direct link, like if someone asks us a question about migrating and we send them the link.

brandonnelson3: Lars was of the mindset of people finding out about this page only when they actually needed to migrate. So I added a check in the install script that basically says, if you already have halyard installed from the debian package, give the user the link to this page.

You can see that change here: https://github.com/spinnaker/halyard/pull/806

dorbin: Ah yes, I remember now. Perf. Approved.

dorbin (approved):

setup/install/halyard_migration.md
dorbin: "Perf" here meaning "perfect." ;)

brandonnelson3: ;) We can always add that link you are suggesting later if it actually be comes a thing we get asked about though. Thanks for the review! This ended up much better IMO than when I first sent it.

PR #507, opened 2018-01-18
fix(artifacts): make binding artifacts clearer

lwander:
https://github.com/spinnaker/deck/pull/4683

### DIFF
--- a/reference/artifacts/in-kubernetes-v2/index.md
+++ b/reference/artifacts/in-kubernetes-v2/index.md
@@ -30,7 +30,7 @@ stage deploys a manifest stored in a GCS bucket:
 Once a manifest has been successfully deployed using a pipeline (either from text
 or an artifact containing text), it is injected back into the pipeline's
 context as an output of the deploy stage. Why this is useful is explained
-[below](#consuming-artifacts-in-manifests), but for now, focus on the distinction between...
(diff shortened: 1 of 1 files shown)


reference/artifacts/in-kubernetes-v2/index.md
```diff
@@ -74,9 +74,9 @@ new version suffix (`-vNNN`). This is critical to supporting immutable
 deployments, as rolling out new ConfigMaps, secrets, or other versioned
 resources should require any manifests that reference them to be updated as
 well. Luckily, Spinnaker makes handling these updates easy as explained
-[below](#consuming-artifacts-in-manifests).
+[below](#binding-artifacts-in-manifests).
 
-# Consuming Artifacts in Manifests
+# Binding Artifacts in Manifests
```
dorbin, line 79: "Binding...in." Is it more correct to say "Binding...to"?

dorbin (approved):
LGTM, thanks!

reference/artifacts/in-kubernetes-v2/index.md
lwander: I think "Binding ... in" is appropriate, since the binding happens "inside" the manifest definition. I think of it as the same as the "bindings in" usage here: https://www.cs.cmu.edu/~rwh/introsml/core/decls.htm

PR #514, opened 2018-01-23
Include ~/.kube when using K8s provider

podlech:


lwander:
Thanks for opening this PR! I think this is a little to k8s specific for the generic "start halyard in docker command". Could you instead move that flag to a comment below explaining why this is needed/helpful?

lwander:
Closing as this section has been rewritten.

PR #515, opened 2018-01-24
Make explicit note that Ubuntu 16 support is not present in Spinnaker mainline releases currently.

ghost:
Per https://github.com/spinnaker/spinnaker/issues/2263 it's somewhat confusing that we label Ubuntu 16.04 support for Halyard but don't make explicit that Spinnaker itself is not yet built and validated for 16.04.

This PR just adds a note to the halyard install docs to make that explicit.

### DIFF
--- a/setup/install/halyard.md
+++ b/setup/install/halyard.md
@@ -26,6 +26,9 @@ There are a few different ways to install Halyard (with more on the way):
 
 ### Ubuntu 14.04/16.04
 
+Note: While Halyard is supported on Ubuntu 16.04 the current release of Spinnaker
+is not.
+
 The following command installs the latest released Halyard version, and will
 prompt the user for some configuration in the process. Generally the default
 answers to each prompt are best.


setup/install/halyard.md
```diff
@@ -26,6 +26,9 @@ There are a few different ways to install Halyard (with more on the way):
 
 ### Ubuntu 14.04/16.04
 
+Note: While Halyard is supported on Ubuntu 16.04 the current release of Spinnaker
+is not.  It's possible to run Spinnaker's nightly builds but that path is not validated.
```
dorbin: Not sure what this 2nd sentence means. Is it necessary, or should we just say "not supported" for now?

ghost: Yup, agreed.

dorbin:
LGTM
For the future, I'd like to find a place to put a statement about what is currently supported, probably [here](https://www.spinnaker.io/setup/install/), plus a mechanism for updating it timely.

sundeer:
I just ran into this which will cost me about an hour.

lwander:
What do you mean? I think the docs are out of date as the most recent (2 or 3) releases support 16.04

sundeer:
Just hit this while trying local install
```
E: Version '0.11.0-20180621044909' for 'spinnaker-igor' was not found
```
Though I was good to go with 16.04 but according to this I need 14.04.
Am I mistaken?

PR #524, opened 2018-01-27
fix(docs): Modify wording of kube provider RBAC

davidxia:
to be more concise and read more smoothly.


davidxia:
I should mention that when I wrote down this RBAC in a previous PR, I was deploying spinnaker with halyard + local debian and only using kube provider v2 on kube 1.8.5. I just tested this RBAC with halyard deploying spinnaker itself onto GKE. I'm not exactly sure what the minimal perms are for kube v1.

lwander (approved):
Nice!

PR #527, opened 2018-01-29
fix(docs): broken swagger-ui link

davidxia:


davidxia:
cc @lwander

lwander (approved):

lwander:
Thanks!

lwander:
I don't think that link is correct, the link checker is failing on it: https://travis-ci.org/spinnaker/spinnaker.github.io/builds/334900930#L2838

davidxia:
Ah, the link is written in a way that makes it work on spinnaker.io [here](https://www.spinnaker.io/reference/api/) but not on github.com. Good catch.

lwander:
Heh the link checker caught it, not me :wink:

PR #532, opened 2018-02-01
feat(provider/kubernetes): v2 deploy guide

lwander:


guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+{%
+  include
+  figure
+  image_path="./deploy-manifest.png"
+%}
+
+> :warning: The regular __Deploy__ stage is used to deploy more opinionated
+> "Server Groups" in Spinnaker, using either another provider, or the
+> Kubernetes V1 provider.
```
dorbin: So you're saying don't select __Deploy__? 

"Don't select the regular __Deploy__ stage, deploys more opinionated "Server Groups," using another provider (including Kubernetes V1)."

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Statically specified manifests
+
+If you know ahead of time what you expect to deploy a certain manifest (even if
```
dorbin: "...deploy *using* a certain manifest..."?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+you don't know what version of your docker image it will run) you can declare
+it directly in the pipeline the manifest specification.
```
dorbin: Are you saying I can declare it in the pipeline by editing the manifest YAML as text, or something like that?

There's a word missing between "pipeline" and "the," maybe? Maybe "in"?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+  "name": "Deploy my manifest",   // human-readable name
+  "type": "deployManifest",       // tells orchestration engine what to run
+  "account": "nudge",             // account (k8s cluster) to deploy to
+  "cloudProvider": "kubernetes",
+  "source": "text",
+  "manifest": {
+                                  // manifest contents go here
+  },
+  "moniker": {                    // used to specify app & cluster for grouping
+                                  // resources in UI
```
dorbin: In the comment for moniker here, I would say " // specifies app & cluster..." rather than "used to specify" because of the other meanings of "used to" (did formerly, or accustomed to).

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+    "app": "xnat",
+    "cluster": "c7",
+  }
+}
+```
+
+### Dynamically specified manifests
```
dorbin: "Specifying Manifests Dynamically"

btw, I see you're using title-caps for some headings and sentence-caps for others. I strongly prefer sentence-caps, like in this one, and spinnaker.io uses an inconsistent mix of both--mostly title-caps. So for now I'm using title caps for everything, but at somepoint I'm going to change them all to sentence-caps. A low-priority effort, but in the mean time I recommend using title-caps everywhere.

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+If you are storing your manifests outside of Spinnaker's pipeline repository,
+or want a single deploy stage to be able to deploy a variety of manifests, you
+can specify your manifest using an [Artifact](/reference/artifacts).
+
+The idea is: artifacts in Spinnaker allow you to reference remote, deployable
+resources. In our case, artifacts will have to reference a text file
```
dorbin: "In our case..." meaning, in the case of the Kubernetes V2 provider?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+(containing a Manifest specification) stored in GitHub, or an object store
+(like GCS).
```
dorbin: remove the comma after GitHub, for clarity

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+There are already some guides describing how to trigger off of changes to these
+artifacts, and supply them to pipelines. See:
+
+* [Consuming GitHub Artifacts](/guides/user/artifacts/github)
+* [Consuming GCS Artifacts](/guides/user/artifacts/gcs)
```
dorbin: Saying "There are already some guides..." colors this document as new, which is true but won't always be true. It's more mundane to say "For more information about triggering based on changes...," but it works better. (Yeah, I also changed "off of" to "based on" cause "off of" is hard-ish to read.

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+{%
+  include
+  figure
+  image_path="./in-artifact.png"
+  caption="Notice that by selecting \"Artifact\" as the __Manifest Source__, we
+  get to pick which upstream artifact to deploy."
+%}
+
+> __☞ Note__: Based on how you configured your artifact accounts, make sure that
```
dorbin: Is "Based on how...accounts," necessary?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+> the __Artifact Account__ field matches an artifact with permission to
+> download your manifest.
+
+Keep in mind that the artifact bound in the upstream stage can match multiple
+incoming artifacts. If instead we had configured it to listen to changes using
+a regex matching `.*\.yml`, it would bind any YAML file that changes in your
+artifact source, and deploy it when it reaches your Deploy stage.
```
dorbin, line 111: This sounds like another warning. Is it another warning? A reminder to make sure to use a specific enough regex to ensure we only pick up the right artifact? Is it necessary?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Overriding Artifacts
+
+In general, when we deploy changes to our infrastructure, the majority of
+changes will come in the form of a new Docker image, or perhaps a feature-flag
+change in a ConfigMap. For this reason, we have first-class mechanisms for
+easily overriding the version a:
```
dorbin: Delete " a" and just put the colon after "version"?

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Docker Image
+* Kubernetes ConfigMap
+* Kubernetes Secret
+
+When any of these objects exists in the pipeline context from an upstream
```
dorbin: s/exists/exist

dorbin: Wait, in the next sentence you say "...automatically try to inject it..."
Is it necessarily one object, or can more than one exist? If only one, then I'd say change the previous sentence to "When one of these objects exists..."

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+stage, Spinnaker will automatically try to inject it into the manifest you're
+deploying. A description of how this works can be read
+[here](/reference/artifacts/in-kubernetes-v2/#binding-artifacts-in-manifests)
+in detail.
+
+To give a quick example, say you trigger your pipeline using a webhook coming
+from a docker registry. At a high level, the event says "Image
```
dorbin: Docker

guides/user/kubernetes-v2/deploy-manifest/index.md
```diff
@@ -0,0 +1,173 @@
...
+  containers:
+  - name: my-container
+    image: gcr.io/my-project/my-image
+# rest of manifest ...
+```
+
+Since the pipeline was triggered by the Docker image changing, the
+orchestration engine will send that artifact along with the manifest to the
+Spinnaker's cloud provider integration service, which, based on the name of the
+docker image, will deploy the following:
```
dorbin: Docker

dorbin (approved):
I'll do an edit on this (and the other V2 docs) in a separate PR.

guides/user/kubernetes-v2/deploy-manifest/index.md
lwander: Good catch, I was missing "by providing"

guides/user/kubernetes-v2/deploy-manifest/index.md
lwander: Yes, will clarify

guides/user/kubernetes-v2/deploy-manifest/index.md
lwander: No, removed

guides/user/kubernetes-v2/deploy-manifest/index.md
lwander: No it's a callout that the stage can be used to deploy multiple manifests. It's not obvious at first, and in some cases is quite helpful once you realize it's possible.

guides/user/kubernetes-v2/deploy-manifest/index.md
lwander: Whoops, meant to have "of" here instead of "a", can I put the colon after "of"?

> For this reason, we have first-class mechanisms for easily overriding the version of:
> ...

dorbin: Yes, but it think it's unncesssary.

guides/user/kubernetes-v2/deploy-manifest/index.md
dorbin: Ah, so you're not  telling people to be careful and avoid too broad a regex, but to be aware that they can do that.

PR #533, opened 2018-02-01
Fix a typo

farcaller:


dorbin:
Good catch, Vladimir. Thanks!

PR #534, opened 2018-02-02
feat(webhooks): document how to use the webhook trigger

lwander:


_data/navigation.yml
```diff
@@ -183,13 +183,15 @@ guides:
             url: /guides/tutorials/codelabs/pubsub-to-appengine/
   - title: User How-Tos
     children:
-      - title: Artifacts
+      - title: Triggers
```
dorbin: Would like a verb here. "Triggering Pipelines"? Configuring Trigger for You Pipelines"? Something describing what the user is doing with triggers.

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
...
+---
+
+{% include toc %}
+
+In order to programatically trigger pipelines you can send a `POST` call to
+Spinnaker against a preconfigured endpoint. This can be used to trigger
+pipelines when a CI job finishes, from a command line, or from a third-party
+system.
+
+> __☞ Note__: There is not a 1:1 mapping between webhooks and pipeline executions.
```
dorbin: I think you can delete the first sentence.

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+To assign an endpoint that must be hit, you can provide a value to the
+__Source__ field as shown here:
+
+{%
+  include
+  figure
+  image_path="./basic-webhook.png"
+%}
+
```
dorbin, line 32: In the image, you've got "demo" in the __Source__ field. What is that? I mean, how does Spinnaker resolve it?

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
+---
+layout: single
+title:  "Triggering on Webhooks"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
```
dorbin, line 9: Re. our discussion about configured external webhooks (like GitHub) versus custom-written webhooks, would it make sense to put this information, about sending a POST, under a subhead, and this first paragraph would instead be an introduction to the idea of some (any) webhook hitting a configured trigger? Or even put this custom POST text in a 
> __Note__:
?

dorbin: "You can configure a webhook to trigger a pipeline. You can do this either by setting up a webhook on some external service (such as GitHub) and configuring a trigger for it in your pipeline, or you can programatically trigger that pipeline by sending a `POST` call against.."

Something like that? Or am I still misconstruing?

guides/user/triggers/webhooks/index.md
dorbin: (answered in person)

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
...
+{%
+  include
+  figure
+  image_path="./basic-webhook.png"
+%}
+
+Notice that in the above image below the __Type__ dropdown, the webhook
+configuration points out that we can hit
+`http://localhost:8084/webhooks/webhook/demo` to trigger the pipeline. The
+endpoint depends on how you've configured your Spinnaker endpoints -- if you're
```
dorbin: Can you link from here to any doc on configuring endpoints?

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
...
+Spinnaker is installed), save that pipeline, and run:
+
+```bash
+curl $ENDPOINT -X POST -H "content-type: application/json" -d "{ }"
+```
+
+### Payload Constraints
+
+If you want to ensure that a webhook only triggers when a certain payload
+arrives, you can provide __Payload Constraints__. These are key/value pairs
```
dorbin: "...you can provide __Payload Contraints__ in the trigger." 

I think it's pretty clear anyway, but it doesn't hurt, since we're configuring the webhook on one end and the trigger on the other, to be clear about where we do each thing.

guides/user/triggers/webhooks/index.md
```diff
@@ -0,0 +1,149 @@
...
+> without providing a default, the pipeline will not trigger if a parameter is
+> not present. The difference between this and the preconditions covered
+> earlier is that when a precondition isn't met, Spinnaker will not even try to
+> run the pipeline. However, when a required parameter doesn't exist, Spinnaker
+> will try and fail to run a pipeline, surfacing a "Failed Execution" in the
+> UI.
+
+## Passing artifacts
+
+If your pipeline requires artifacts (e.g. a Kubernetes manifest file stored in
```
dorbin: Per [Developer Docs Style](https://developers.google.com/style/word-list#letter-E), use "For example" rather than "e.g."

dorbin (approved):
More good doc, Thanks!

guides/user/triggers/webhooks/index.md
lwander: The configuration for a GitHub trigger is maybe a bad example because it is configured entirely differently, but I can point out that these triggers don't have to come from your command line

lwander: rewritten and made the third-party vs your own payload distinction a little clearer

PR #535, opened 2018-02-02
Point out to operators to replace 'https' with 'http' in the OAuth callback if appropriate

demobox:


lwander (approved):

demobox:
> lwander approved these changes

Thanks, Lars!

PR #539, opened 2018-02-05
feat(artifacts): document artifact types

lwander:


reference/artifacts/types/docker-image/index.md
```diff
@@ -0,0 +1,35 @@
...
+[GCR](https://gcr.io), or [Docker Hub](https://index.docker.io). They can be
+deployed to Kubernetes or App Engine, and generally trigger pipelines from
+notifications sent by their registry.
+
+## Fields
+
+| Field | Explanation |
+|-|-----------|
+| `type` | Always `docker/image`. |
+| `reference` | The fully-qualified image name, including version and registry. `gcr.io/project/my-image:v1.2` is an example. This can be done either by tag or image digest. |
```
dorbin: And this example is the former, right? Would it work to put this example in parens after "by tag" and and example digest in parens after "or image digest"?

reference/artifacts/types/docker-image/index.md
```diff
@@ -0,0 +1,35 @@
...
+| `name` | The image's name (typically the registry and repository) without the image's version. `gcr.io/project/my-image` is an example. |
+| `version` | The image's tag or digest. |
+| `location` | N/A |
```
dorbin: Just wondering&mdash;I'm sure there's a good reason, but why do we need all three fields: reference, name, and version? Is it because if we use the digest for "reference," we can't easily see the name or version?

reference/artifacts/types/gcs-object/index.md
```diff
@@ -0,0 +1,34 @@
...
+sidebar:
+  nav: reference
+---
+
+{% include toc %}
+
+GCS object artifacts are stored in [GCS
+Buckets](https://cloud.google.com/storage/). They are generally consumed by
+stages that read configuration from text files, such as a Deploy Manifest or
+Appengine Deploy stage.
```
dorbin: "App Engine" ?

reference/artifacts/types/overview/index.md
```diff
@@ -0,0 +1,27 @@
+---
+layout: single
+title:  "Types of Artifacts"
+sidebar:
+  nav: reference
+---
+
+While an artifact can reference any remote, deployable resource, there are
+certain types of artifacts that we have first-class support for in the form of:
```
dorbin: "While an artifact can reference any remote, deployable resource, we have first-class support for certain types of artifacts, in the form of..."

[list]

reference/artifacts/types/overview/index.md
```diff
@@ -0,0 +1,27 @@
...
+  from GCR images.
+
+* Credentials for downloading artifacts. For example, a GitHub access token to
+  read repository contents.
+
+* Integrations with stages that require certain types of artifacts. For
+  example, App Engine can only deploy a Docker image as a custom runtime, never
+  an AMI.
+
+The pages in the section serve to document the format Spinnaker expects these
```
dorbin: "The pages in *this* section describe the format Spinnaker expects for these artifacts, to make custom integrations easier."

reference/artifacts/types/overview/index.md
```diff
@@ -0,0 +1,27 @@
...
+artifacts in, to make custom integrations with them easier. We recommend
+reading the [artifact format](/reference/artifacts/#format) first.
+
+These pages _do not_ give concrete guides or practices for automatically
+sending or injesting artifacts in pipelines. Those can be found in the
+[guides](/guides) pages.
```
dorbin, line 27: Maybe a more specfic link within /guides?

dorbin (approved):
You're basically _defining_ each of these artifact types, right? But for some, you're  starting off with informaition about them, like where they're stored.  

For Kubernetes Object, though, you start off with what it is. Could you do that for all of them.

reference/artifacts/types/docker-image/index.md
lwander: You can't always derive reference from the other fields, and even in cases where you can, there is no artifact-agnostic way to derive version and name from reference.

reference/artifacts/types/docker-image/index.md
lwander: Good idea

reference/artifacts/types/overview/index.md
lwander: Unfortunately there is none yet, I will file an issue to update this with a better section once there is one.

lwander: https://github.com/spinnaker/spinnaker.github.io/issues/540

lwander:
Updated all artifact descriptions

PR #541, opened 2018-02-06
fix(nav): architecture heading links to overview

lwander:
f4szxhbflcd

Usually those headings don't link at all

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -264,7 +264,6 @@ reference:
       - title: Component Sizing
         url: /reference/halyard/component-sizing
   - title: Architecture
-    url: /reference/architecture/
     children:
       - title: Overview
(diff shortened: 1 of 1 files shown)


dorbin (commented):
Good catch, Lars. When I re-did the Setup tab, I set all of these top-level navs as collapsed by default. But not all tabs are done this way. Yet. There's an argument to be made for either choice. It definitely tells you more about what's in a section if the first level is expanded by default. But I will point out that on my laptop, the nav for Reference spills below the fold--don't even see the Architecture section.

lwander:
I think the trouble with having the top-level nav be clickable is that it breaks the zippy behavior (you can't expand/collapse without clicking).

dorbin (approved):

PR #545, opened 2018-02-08
fix(artifacts): fix github artifact setup

lwander:


setup/artifacts/github.md
```diff
@@ -39,6 +41,7 @@ First, make sure that artifact support is enabled:
 
 ```bash
 hal config features edit --artifacts true
+hal config artifact github enable
```
dorbin: Above, it's "First, make sure that artifact support is enabled" which would still be `hal config features edit --artifacts true` right?  Should there be an additional step here for enabling github?

dorbin (approved):
lgtm

PR #552, opened 2018-02-12
fix(artifacts): add more explicit information for pulling GitHub file…

clanesf:
…s as artifacts

Added an example for GitHub Enterprise and clarified that this reference field points to the GitHub API.


lwander (approved):
Thank you!

Issue #556, opened 2018-02-13
Local Debian install is not supported in the docker image


snstanton:
Page: /setup/install/environment/

Feedback:

The Debian docker image supplied by the halyard setup does not work as a local installation environment.  This should be mentioned in this document, as it is confusing for a first time installation to run in to the error message that Debian is not supported.

dorbin:
Thanks, Scott. Would you like to add a note to that page?

lwander:
Closed by #557

PR #557, opened 2018-02-13
Local Debian install is not supported in the docker image (#556)

snstanton:


lwander (approved):
Thank you! We appreciate it

PR #558, opened 2018-02-13
chore(css): round box comment sections

lwander:
I'll be breaking these out by warning, comment, etc... later to allow us
to insert "Warning! Alpha feature..." for all k8s v2 pages

makecommentsnicer

### DIFF
--- a/_sass/_page.scss
+++ b/_sass/_page.scss
@@ -61,6 +61,12 @@
     font-size: 0.8rem;
   }
 
+  blockquote {
+    p {
+      margin: 1em;
+    }
+  }
+
   /* paragraph indents */
   p {
     margin: 0 0 $indent-var;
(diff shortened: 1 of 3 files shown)


ghost (approved):

_sass/_page.scss
```diff
@@ -61,6 +61,12 @@
     font-size: 0.8rem;
   }
 
+  blockquote {
+    p {
+      margin: 1em 1em 1em;
```
ghost: v minor nit: shorthand `margin: 1em;` should have the same effect here.

lwander: Isn't that the same as `margin: 1em 1em 1em 1em`? I don't want the left margin to be `1em`

ghost: `margin: 1em 1em 1em;` is also the same as `margin: 1em 1em 1em 1em;`. when you drop the left margin value it takes the value from the right margin unless you override it elsewhere.

lwander: woah

PR #561, opened 2018-02-14
fix(codelabs): Update create instance parameters for GCE codelab

ezimanyi:
Publishing a new image for the GCE codelab based on spinnaker 1.5.4,
which requires a change to the parameters we use to create an instance
from this image.


ezimanyi:
@lwander @dorbin : Mind taking a quick look at this change?

dorbin:
LGTM

lwander (approved):

PR #568, opened 2018-02-15
feat(artifacts): add warning about GCR messages for app engine

ghost:


setup/triggers/google.md
```diff
@@ -92,6 +92,16 @@ material](https://cloud.google.com/storage/docs/pubsub-notifications).
...
 ### Receiving messages from Google Container Registry (GCR)
 
+{% include
+   warning
+   content="If you are planning to use artifacts from GCR messages to
+   trigger a Google App Engine deployment then you will need to use a
+   custom message template. The template will need to copy an image's
+   tag into the artifact's reference and version fields.  App Engine
+   does not support deploying from an image's digest, which the GCR
+   message type extracts by default."
```
dorbin: "The GCR message type extracts the image digest by default, but Google App Engine doesn't support deploying from a digest. So if you're using artifacts from GCR messages to trigger a GAE deployment, you need to use a custom message template. The template must copy the image's tag into the artifact's reference and version fields. "

ghost: agreed and fixed :+1:

dorbin (approved):

lwander (approved):

PR #571, opened 2018-02-16
feat(kubernetes): add manifest docs to index

lwander:
We should wait to merge until 1.6 is released


lwander:
https://github.com/spinnaker/spinnaker/issues/2231

dorbin (approved):
So that's it then? Out the door? 
Great job on the doc!

As I said, I'll do an editing run on the whole collection, but this doesn't need to wait on that, because alpha.

lwander:
Merging since 1.6 is out

PR #573, opened 2018-02-16
fix(codelabs): corrects erroneous registry name

dorbin:


skim1420:
The screenshot just above the text is incorrect as well. Either update the screenshot (preferred) or remove it altogether.

dorbin:
Commented it out. I think the screenshot should be in, but I'm not going to be able to work through it to generate the shot for a while.

skim1420 (approved):

PR #574, opened 2018-02-16
fix(build): add Gemfile.lock to repo

ghost:
Include a Gemfile.lock so that we have a record of the explicit dependency versions needed for a working build of the site.


ghost:
@dorbin the travis build is fixed and this is ready for your review.  The fix for travis was to run `rake test` using the more verbose `bundle exec rake test` which I believe provides some environment changes to deal with installing from the Gemfile.lock.

dorbin (approved):
LGTM. Thanks for the fix!

PR #575, opened 2018-02-20
feat(provider/kubernetes): document remaining resource types

lwander:


reference/providers/kubernetes-v2/index.md
```diff
@@ -200,5 +200,20 @@ be rendered as well.
...
+| __Resource__ | _Deploy_ | _Delete_ | Versioned | Stability |
+|-|:-:|:-:|:-:|-|
+| __`NetworkPolicy`__ | Yes | Yes | No | Automatically "stable". |
+
+## ConfigMaps, Secrets
+
+| __Resource__ | _Deploy_ | _Delete_ | Versioned | Stability |
+|-|:-:|:-:|:-:|-|
+| __`ConfigMap`__ | Yes | Yes | Yes | Automatically "stable". |
+| __`Secret`__ | Yes | Yes | Yes | Automatically "stable". |
```
dorbin: I wonder if it might be better, instead of putting "stable" in double-quotes, linking it to the __Stability__ heading under __Resource Management Policies__, or to __Accomodating Level-Based Deployments__, where you have your statement, "A deployed manifest is considered stable when..." In fact, would it make sense to move that definition of stability to under the aforementioned Stability heading.

dorbin: BTW, "Accomodating" in that heading is misspelled.

dorbin (approved):

reference/providers/kubernetes-v2/index.md
lwander: SGTM

PR #580, opened 2018-02-22
Update GKE Quickstart to download latest kubectl

adamresson:
Ran into an issue with the halyard Quickstart for GKE docs where I was unable to deploy:
```
! ERROR Could not parse connection information from:
()

- Failed to deploy Spinnaker.
```

Turns out it was due to an older client version of kubectl. When I went back to see how the quickstart guide suggested I install kubectl, i noticed the docs have it hardcoded to 1.6.1. Updating docs so that recommendation is to install latest stable, as recommended on the kubectl install page.


dorbin:
LGTM
@lwander?

lwander (approved):
Thanks!

PR #582, opened 2018-02-22
Fix long line in GKE Quickstart

adamresson:
Extra long line from #580 looks terrible in docs page (https://www.spinnaker.io/setup/quickstart/halyard-gke/). Oops!


dorbin (commented):
@adamresson , looks like you accidentally checked in a binary file along with this change?

adamresson:
@dorbin fixed

PR #583, opened 2018-02-22
Fix receive spelling

bvold:


dorbin (approved):
Thanks for catching those, @bvold

PR #587, opened 2018-02-23
feat(kubernetes): v2 automated rollbacks guide

lwander:


guides/user/kubernetes-v2/automated-rollbacks/index.md
```diff
@@ -0,0 +1,96 @@
...
+{% include alpha version="1.6" %}
+
+{% include toc %}
+
+There are clear benefits to having your stored manifests (in Git, GCS, etc...)
+match as closely as possible to what's running in your cluster: you have
+auditability, versioned changes, and a high confidence in what code you have
+running. However, in some cases, pushing changes from someplace like GitHub
+through your build system all the way production can simply take too long, and
+an escape hatch is needed when something has gone seriously wrong.
```
dorbin: Do you mean something can go wrong _because of_ the time lag?

Can you give an example of something going wrong?

guides/user/kubernetes-v2/automated-rollbacks/index.md
```diff
@@ -0,0 +1,96 @@
...
+{% include
+   figure
+   image_path="./v005.png"
+%}
+
+## Automated Rollbacks
+
+You can also configure automated rollbacks inside of Spinnaker pipelines. These
+stages are best configured when other stages or branches fail, indicating that
+something has gone wrong in your rollout.
```
dorbin: So an upstream stage fails, and that failure triggers an auto-rollback stage?

guides/user/kubernetes-v2/automated-rollbacks/index.md
```diff
@@ -0,0 +1,96 @@
...
+following state:
+
+```
+nginx-deployment-2d8178b77 (Revision 5) # active
+nginx-deployment-7bdd110f7 (Revision 4) 
+nginx-deployment-0b13cc8c1 (Revision 1) 
+```
+
+And roll back by "1" revision, (Revision 3) will be active again. Rollback by
+"2" revisions and (Revision 1) will be active again.
```
dorbin: Rev 5 here is "active." Does that mean that's the deployed version, and my pipeline is trying to deploy, say, Revision 6? If so, why does the rollback by 1 mean rev 3 is now active? Why not 4?

guides/user/kubernetes-v2/automated-rollbacks/index.md
```diff
@@ -0,0 +1,96 @@
...
+
+> Keep in mind that Kubernetes will implicitly rollforward the old
+> configuration, creating (Revision 6) in both cases.
```
dorbin: Right, just to make sure I understand, not necessarily that you need to clarify for reader...

Roll back N rolls back to N-1 (for example), but _calls it_ N+1, because now it's a new revision? But it's identical to N-1?

guides/user/kubernetes-v2/automated-rollbacks/index.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Parameterized Rollbacks
+
+It's worth mentioning that you can parameterize the target resource to
+rollback. It can point to something specified using pipeline parameters, and
+upstream deploy stage, or another stage's outputs. See more details in the
+[pipeline expressions guide](/guides/user/pipeline-expressions).
```
dorbin: Confused by this paragraph. Is this it?
"You can use pipeline parameters to specify the target resource to roll back, based on an upstream deploy stage or another stage's output." (Or to roll back _to_?)

dorbin (commented):
Questions before I approve; I just want to make sure I understand it. Thanks!

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: I mean that once something has gone wrong, the time lag might be too long to be acceptable. There are always ways around this, but (for example) to build our clouddriver docker image, validate it, and (hypothetically) push it to production would take roughly 30 minutes. Simply clicking "rollback" takes less than 30 seconds.

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: Yup! I think a guide for configuring stages that run on "failure" merits another guide.

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: It means it's the revision serving traffic.

Whoops, the "revision 3" is a typo

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: Yup that's right, it's a Kubernetes quirk

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: You could do either, but in this case it's referring the former: we are saying "roll back this service", not "roll back this service to this revision" using the pipeline parameters. Let me know if that can be worded better

guides/user/kubernetes-v2/automated-rollbacks/index.md
dorbin: Ah, I'm getting it now. Maybe an additional sentence at the end: "Automated rollback is that escape hatch."

guides/user/kubernetes-v2/automated-rollbacks/index.md
dorbin: "These stages are best configured to run when other stages..."

guides/user/kubernetes-v2/automated-rollbacks/index.md
dorbin: "...that you can parameterize which resource to roll back."

dorbin (approved):

guides/user/kubernetes-v2/automated-rollbacks/index.md
dorbin: BTW, I like "roll back" as a verb and "rollback" as noun. I'll run that by the editoria discuss group here.

lwander: SGTM, I'll switch the wording here

guides/user/kubernetes-v2/automated-rollbacks/index.md
lwander: https://github.com/spinnaker/spinnaker.github.io/issues/598

PR #591, opened 2018-02-23
feat(kubernetes): manifest parameterization guide

lwander:


guides/user/kubernetes-v2/parameterize-manifests/index.md
```diff
@@ -0,0 +1,60 @@
...
+title:  "Parameterize Kubernetes Manifests"
+sidebar:
+  nav: guides
+---
+
+{% include alpha version="1.6" %}
+
+{% include toc %}
+
+It is possible to have Spinnaker inject context from the currently executing
```
dorbin: "Spinnaker can inject..."

guides/user/kubernetes-v2/parameterize-manifests/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+```yaml
+# ... other keys
+metadata:
+  namespace: '${ parameters.namespace }'
+# other keys ...
+```
+
+> Keep in mind this manifest can be stored in the pipeline, or in an external
+> store such as GitHub. Read more
```
dorbin: You could delete the "Read more here" sentence, and instead link from "can be stored in the...or in an external store"

guides/user/kubernetes-v2/parameterize-manifests/index.md
```diff
@@ -0,0 +1,60 @@
...
+   figure
+   image_path="./parameter.png"
+%}
+
+> See more details on how to provide parameters to piplines programatically in 
+> the [wehbooks](/guides/user/triggers/webhooks) page.
+
+## Configure your Manifest
+
+We want the user-supplied parameter to specify the manifest's namespace. Edit
```
dorbin: Maybe make it clearer that this is an example.

"In this scenario, we're using a parameter to specficy the manifest's namespace."

Or even "We want _this_ user-supplied parameter to..."

dorbin: (Consider the case where the careless reader skips the intro sentence of this article and goes right for the meat of the instructions.)

dorbin (approved):
Looks good, Lars, thanks.

PR #612, opened 2018-03-09
fix(text): set body font color to black

### LINKED ISSUE #610: Contrast ratio of spinnaker.io body text possibly too low
TLDR: I'd like to propose that we move to a thicker font on spinnaker.io in order to increase the text contrast.

Main body text on spinnaker.io is quite thin, resulting in a lack of contrast between it and the page's white background.  Here is a screenshot that demonstrates how text looks on the site currently and how it appears with a thicker font:

skinny_vs_thick_contrast_ratio

The difference in contrast is quite clear.  I've personally experienced a bit of eye strain while reading the site and I believe that the contrast ratio is quite possibly below minimum accessibility guidelines of 4.5:1.  It's difficult to test this accurately because the actual text color value is above accessibility guidelines but the thinness of the font reduces it.

This is an issue I've worked around on my local machine using a stylesheet override but I'd like to modify the site to use a thicker font or darker colour by default.

ghost:
fixes https://github.com/spinnaker/spinnaker.github.io/issues/610

Before:

before_font_color_change

After:

after_font_color_change

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -48,6 +48,7 @@ $type-size-8                : 0.625em !default;  // ~10px
    Colors
    ========================================================================== */
 
+$black                      : #000 !default;
 $gray                       : #7a8288 !default;
 $dark-gray                  : mix(#000, $gray, 40%) !default;
 $darker-gray                : mix(#000, $gray, 60%) !default;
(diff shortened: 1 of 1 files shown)


dorbin (approved):
Thanks, Scott!

PR #622, opened 2018-03-16
Add note to comparisons about data typing

dotdotdotpaul:
Took me forever to realize that  #stage('Validation')['status'] == 'SUCCEEDED' wasn't working because the status is an enum internally, not a string (even though in the JSON it's represented as a string.  Solution was to add .toString() to it.  This seemed like such an inobvious solution/problem that it warranted a note in the docs.


lwander (approved):
Thank you for calling this out!

Could you add an explicit example on the line below similar to what you've described in your commit message? Just to make this very clear.

dotdotdotpaul:
Example added!  :smile:

lwander:
Thanks!

PR #624, opened 2018-03-18
Fix a typo

jianglai:
$PROJECT should be $GCP_PROJECT.


dorbin:
Good catch, Lai, thanks!

dorbin (approved):

PR #625, opened 2018-03-19
chore(install): move manifest-based install helper

lwander:
After chatting with @emjburns about where the AWS install help should go we're moving this into a dedicated kubernetes subfolder.


duftler (approved):
LGTM.

PR #631, opened 2018-03-21
fix(gke/install): fix endpoint for cloud shell

lwander:
The localhost baseurl breaks when in cloudshell (doesn't forward to
localhost)

### DIFF
--- a/downloads/kubernetes/quick-install.yml
+++ b/downloads/kubernetes/quick-install.yml
@@ -230,7 +230,7 @@ data:
         apiSecurity:
           ssl:
             enabled: false
-          overrideBaseUrl: http://localhost:9000/gate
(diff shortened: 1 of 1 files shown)


downloads/kubernetes/quick-install.yml
```diff
@@ -230,7 +230,7 @@ data:
         apiSecurity:
           ssl:
             enabled: false
-          overrideBaseUrl: http://localhost:9000/gate
+          overrideBaseUrl: /gate
```
jtk54, line 233: Does this work when you're not running on cloud shell?

lwander: yes

jtk54 (approved):

PR #635, opened 2018-03-27
fix(docs): Update RBAC permissions required for Spinnaker

ezimanyi:


ezimanyi:
Two things:
-Noticed that I was getting errors in the apiserver log that Spinnaker did not have create permission on "pods/portforward"
-Some of the existing entries have curly quotes, which leads to a silently invalid config file when copy-pasted

dorbin:
Good catch on those curly quotes, @ezimanyi !

lwander (approved):

PR #638, opened 2018-03-29
docs(codelabs): updated screen shot of Jenkins global security settin…

dorbin:
…gs to match instructions


lwander (approved):
Looks safer!

PR #640, opened 2018-03-29
Docs(Reference): removes dependency diagram, to rely only on table instead

dorbin:


duftler (approved):
LGTM.

ttomsu:
Booooooo - I reference the visual diagram when introducing the architecture to new people. Please bring it back!!

PR #649, opened 2018-04-03
fix(quick-install): allow hal config to be edited after quick install deployed

ghost:


lwander (approved):
Nice!

downloads/kubernetes/quick-install.yml
```diff
@@ -45,6 +45,11 @@ spec:
         # todo - make :stable or digest of :stable
         image: gcr.io/spinnaker-marketplace/halyard:0.43.0-20180321030509
         imagePullPolicy: Always
+        command:
+        - /bin/sh
+        args:
+        - -c
+        - "cp -R /root/staging/.hal /root/.hal && /opt/halyard/bin/halyard"
```
lwander: Can you add a yaml comment explaining why we do this?

PR #651, opened 2018-04-04
Document S3 Artifact type

benjaminws:
Light documentation to support https://github.com/spinnaker/spinnaker/issues/2595

### DIFF
--- a/reference/artifacts/index.md
+++ b/reference/artifacts/index.md
@@ -109,7 +109,7 @@ within pipelines, or produced by pipelines.
 ```
 
 ```js
-// A S3 object
+// An S3 object
 {
   "type": "s3/object",
   "name": "s3://bucket/file.json",
--- a/reference/artifacts/index.md
+++ b/reference/artifacts/index.md
@@ -112,9 +112,7 @@ within pipelines, or produced by pipelines.
 // A S3 object
 {
   "type": "s3/object",
-  "reference": "s3://bucket/file.json#1522839181",
   "name": "s3://bucket/file.json",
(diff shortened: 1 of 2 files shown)


reference/artifacts/index.md
```diff
@@ -107,3 +107,14 @@ within pipelines, or produced by pipelines.
   "location": "us-central1"
 }
 ```
+
+```js
+// A S3 object
```
dorbin: s/A/An

reference/artifacts/types/s3-object/index.md
```diff
@@ -0,0 +1,31 @@
...
+| `name` | Name/Path of the object. |
+| `location` | The region where your bucket was created |
+
+## Example
+
+```js
+{
+  "type": "s3/object",
+  "name": "s3://bucket/file.json",
+  "location": "us-east-1"
```
dorbin, line 29: In the example you added to /reference/artifacts/index.md you included `reference` and `version` fields. Should those be added here, and to the list of fields above?

dorbin (approved):
Thanks for the doc, Benjamin!

reference/artifacts/types/s3-object/index.md
benjaminws: So it's actually the other way around. S3 doesn't version the files the same way, so those aren't used here. Updated in the example

reference/artifacts/index.md
benjaminws: Fixed!

lwander (approved):
Thanks!

PR #655, opened 2018-04-04
Makes the feedback contents include a link to the page, instead of ju…

ttomsu:
…st the text of the path.

pmw93pec6v8


dorbin (approved):
Thanks, T!

PR #658, opened 2018-04-06
fix(docs/quickstarts): remove nav links to the full-install quickstart; replace 'quickstart' with 'try out' in the full install docs.

avh121:


_data/navigation.yml
```diff
@@ -29,8 +29,6 @@ setup:
     children:
       - title: Demo Install
```
demobox: Do we still need children under this menu item if there's only one? I.e. would it be possible to just make the "Quickstarts" text take you straight to the "here's a list of things to try out" page?

setup/quickstart/halyard-gke-public/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: single
-title:  "Public Spinnaker on GKE"
+title:  "Try out public Spinnaker on GKE"
```
demobox, line 3: [minor and possibly to be addressed in a separate issue] Do we feel people are likely to know what "public Spinnaker" means here?

setup/quickstart/halyard.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: single
-title:  "Quick Full-Installations Using Halyard"
+title:  "Try out full installations using Halyard"
```
demobox, line 3: These are not "full installations" anymore? I'm assuming this would be the single page that clicking on "Quickstarts" would take you to..?

demobox (changes requested):

setup/quickstart/halyard.md
avh121: No, these are still the full installation pages. I removed the nav links to them but didn't delete them. This way we can use some of this content for re-doing the non-quickstart setup docs, and people who have this bookmarked will still have access for the moment.

setup/quickstart/halyard-gke-public/index.md
avh121: Honestly, I'm not sure what "public Spinnaker" means, so probably not :) I didn't address it here because I wanted to keep this PR focused, but I agree that we should probably pick a clearer term.

_data/navigation.yml
avh121: Yep, I was just going to do this in a follow-up PR :) Changed it here though.

setup/quickstart/halyard.md
demobox: > This way we can use some of this content for re-doing the non-quickstart setup docs

Ah, OK, thanks for explaining. Given that part of the point of this restructuring is to give people _fewer_ options for potentially doing "full installations" of Spinnaker, do you think it would make sense to add a header to each one of these to explain that the **recommended** way of doing a full Spinnaker installation is to follow the "normal" setup guide?

avh121: Yep, that seems reasonable; done.

demobox: > Yep, that seems reasonable; done.

Thanks! Curious if we have any standard way of marking particular sections as deprecated/no longer recommended, but we can always take care of that in a follow-up.

/cc @dorbin

demobox (approved):

setup/quickstart/halyard-gce.md
```diff
@@ -1,13 +1,15 @@
...
 layout: single
-title:  "Halyard on GCE Quickstart"
+title:  "Try out Halyard on GCE"
 sidebar:
   nav: setup
 ---
 
 {% include toc %}
 
+Note: we recommend that you install Spinnaker following the [standard setup directions](/setup/) rather than the guides listed here.
```
dorbin: I don't know if "guides," plural, works well here. Because now they're not listed. 

"We recommend you install...rather this using this guide, which is just a set of commands to get your Spinnaker up and running on GCS and GCE."

I think what we might decide to do is delete all these and redirect to /setup/install/.  @demobox WDYT?

Also, @nhayes you can indent this as a "note" using `> ` at the start of each line.

setup/quickstart/halyard.md
dorbin: I think our standard way is to delete and redirect.

dorbin (approved):
Thanks, Nora!

lwander:
@duftler fyi

Why were these removed?

dorbin:
@demobox 

Part of our broader, ongoing effort to reorg all our getting started stuff. These are not considered "quickstarts" for the purposes of getting started. 

What's there now is a set of links to the Cloud Launcher and equivalents. These are just to get Spinnaker running for tire-kicking purposes.

What's removed are the Halyard-GKE solutions, which we are thinking about putting somehwere else.

lwander:
I think in the future if we are planning on moving docs we should do that in the same PR. These were useful, and w/o a bookmark or link are now very hard to find.

PR #678, opened 2018-04-12
fix(install/halyard): mount .hal under spinnaker user

lwander:
Don't merge until https://github.com/spinnaker/halyard/pull/911 is merged and pushed to `:stable`


lwander:
@dorbin FYI

PR #679, opened 2018-04-12
Sent

avh121:


reference/providers/gce.md
```diff
@@ -94,16 +94,16 @@ receives traffic.
...
 Performs a [hard reset](https://cloud.google.com/compute/docs/instances/restarting-an-instance) on an instance.
 
-### Terminate Instance
+### Terminate instance
 
 [Deletes](https://cloud.google.com/compute/docs/instances/stopping-or-deleting-an-instance#delete_an_instance) an
 instance. In most cases, the managed instance group will provision a new instance to replace the terminated instance.
 
-### Terminate Instance and Shrink Server Group
+### Terminate instance and shrink Server Group
```
dorbin, line 106: Wondering if "server group" should  be lower case. There's no entry for it in the [Dev Docs Style Guide](https://developers.google.com/style/). 

Just search c.g.c, and there are three [results](https://cloud.google.com/s/results/?q=%22server+group%22&p=%2F) for "server group," all lower case.

dorbin (approved):
Looks great! Thanks, Nora!

reference/providers/gce.md
avh121: I left it uppercase because we had talked about treating load balancer as a proper noun when I was updating some of the other docs, and I assumed that server group fell into the same category. If I change "server group" to be lower case here, should I change "load balancer" and "security group" to be lower case as well?

dorbin: Ah yes, I remember. You're right. From the Spinnaker perspective, they're equivalent. Keep upper.

Thanks!

PR #680, opened 2018-04-13
fix(docs): replace spinnaker blue/green with red/black

tomaslin:
there is no such thing as blue/green in spinnaker


lwander (approved):

skim1420:
How about we compromise and do black & blue?

PR #682, opened 2018-04-16
Correct BYO-oauth provider instructions

ttomsu:
https://www.spinnaker.io/reference/halyard/custom/#custom-profiles says to put it in `profiles`


dorbin (approved):
LGTM, thanks!

lwander (approved):

PR #696, opened 2018-04-20
Fix breakline in table at monitoring document

dtan4:
The bottom table style in [monitoring doc](https://www.spinnaker.io/reference/monitoring/) is broken.
Using `<br>` tag fixes this.

### Before

image

### After

image


dorbin (approved):
Thanks, Daisuke!

PR #703, opened 2018-04-23
fix(auth): link in documentation prompts for atlassian login

### LINKED ISSUE #2691: Link in documentation prompting for Atlassian login
A link to info on LDAP search filters is prompting for Atlassian login

Cloud Provider: N/A

Environment: N/A

Feature Area: N/A

Description:
In the last paragraph of https://www.spinnaker.io/setup/security/authentication/ldap/#configure-ldap-using-halyard there is a link to an Atlassian page that prompts for login.

_You can also use ldap.userSearchBase and ldap.userSearchFilter if the simpler ldap.userDnPattern does not match what your organization uses for userDns. We don’t explore this use case here, but you can read up more on LDAP search filters here._

Link: https://confluence.atlassian.com/display/DEV/How+to+write+LDAP+search+filters

Expected to get info on LDAP filters, but it seems like I would need an Atlassian account to read these docs.

ghost:
fixes https://github.com/spinnaker/spinnaker/issues/2691


dorbin (approved):
Good catch, Scott!

PR #705, opened 2018-04-25
fix(docs): Fixes broken link to the ContextParameterProcessor

antonio-osorio:
ContextParameterProcessor was migrated from groovy to java and the reference in the doc was not updated.


ttomsu (approved):

ttomsu:
Thank you for the fix!

PR #713, opened 2018-04-30
feat(docs): enable collapsible content sections

ghost:
This PR introduces a mechanism for collapsible content sections.  When the page loads, the script searches for elements with a `begin-collapsible-section` class.  If found the script then searches forward for an `end-collapsible-section` element.  If found then all content between those two elements is wrapped in a collapsible div and hidden.

This is definitely not elegant but works around the design of markdown where wrapping content in a `<div>` disables markdown parsing of that content.

@dorbin I've added the first collapsed section: The Local Debian setup instructions are automatically collapsed.  Is this desirable or would you prefer I leave it out from this PR and you can add collapsible sections going forward?

### DIFF
--- a/_includes/collapsible_content_script.html
+++ b/_includes/collapsible_content_script.html
@@ -1,0 +1,33 @@
+<script>
+  $(document).ready(function() {
+    $('.begin-collapsible-section').each(function(i, el) {
+      var $start = $(el);
+      var $parent = $start.parent();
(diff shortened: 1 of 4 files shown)


dorbin:
@sbwsg , I think it's ok to leave the page-specific uses of it in this PR. It serves as a good example of how to do it.

dorbin (approved):

PR #721, opened 2018-05-01
docs(authentication): adds verification and troubleshooting to ssl config

dorbin:


setup/security/authentication/index.md
```diff
@@ -3,6 +3,8 @@ layout: single
 title:  "Authentication"
 sidebar:
   nav: setup
+redirect_from: /setup/security/authentication/network-arch/
```
stewchen, line 6: What does this do? Network Architecture links should also be removed from this page

stewchen (approved):

setup/security/authentication/index.md
dorbin: That's in case anyone has a bookmark or direct link to that page (unlikely but possible).  User gets redirected to this page instead of a 404.

PR #729, opened 2018-05-03
fix(docs): code font nested in list too small

ghost:
This normalizes the font size of nested code blocks to match that of top-level code blocks.


dorbin (approved):
Thanks for finding the fix, Scott.

PR #732, opened 2018-05-04
fix(docs/codelabs): clarify how to deploy the latest tag from dev to prod.

### LINKED ISSUE #376: Prod server not updated with new image
Page: /guides/tutorials/codelabs/kubernetes-source-to-prod/

Feedback:
Great tutorial thanks.

I have a problem that updated code only makes it to the dev server group. When I update code, the deployment pipeline is successful all through to prod, but the prod deployment simply re-deploys the first tag defined.

In my example below, the first deployment is v18

Following the instructions, in the deploy to dev stage, for container we select "Tag resolved at runtime", v18 is deployed and in the first run, v18 is deployed to prod. Now I update the code to v19, and this is deployed to dev. However, v18 is re-deployed to prod.

When setting up the deploy to prod stage, the "tag resolved at runtime" option is not available and we select a tag (v18). It is this tag that is always deployed. In the Deploy stage, surely we should be deploying the result of the Find Image from Cluster stage, the result of which does correctly identify the last deployed tag (v19). 

It seems there is a stage or something missing as it seems the tutorial will always deploy the same tag to prod, as that is what is defined. What am I missing here? 

This is my first post so let me know what further information/screenshots are required. I have followed the tutorial fully, kubernetes pod dev info shows v19 and correctly increments (v20, v21 etc), the prod pods all show v18 for deployments

avh121:
Fixes #376.


dorbin (approved):
LGTM, if it LGT @lwander 

Thanks!

lwander (approved):
LGTM!

Issue #697, opened 2018-04-20
Zero-to-one: Install Halyard


dorbin:
Re-write step 1. [Install Halyard](https://www.spinnaker.io/setup/install/halyard/)

dorbin:
#733

PR #733, opened 2018-05-04
docs(Halyard): improvements to Install Halyard

dorbin:


dorbin:
#697

lwander (approved):
Nice!

dorbin:
@joonlim, for you to test

joonlim:
Tested "Install Halyard on Docker" and "Update Halyard on Docker" flows and they work well.

One suggestion I have is that we may want to tell users how to run halyard in detached mode in case they are not interested in the logs:
```
docker run -p 8084:8084 -p 9000:9000 \
    --name halyard --rm \
    -v ~/.hal:/home/spinnaker/.hal \
    -d \
    gcr.io/spinnaker-marketplace/halyard:stable
```
Personally, if I were installing halyard for the first time, I would be wondering why I would care about the logs at all.

lwander:
@joonlim that's fair - we can always ask them to restart w/ logging to help them fix an issue

setup/install/halyard.md
```diff
@@ -70,7 +70,12 @@ your container:
 
    You can interact with Halyard from here.
 
-1. Run `. ~/.bashrc` to enable command completion.
+1. Run the following commands to enable command completion:
+
+   ```bash
+   hal --print-bash-completion > ~/hal_bash_completion
+   source ~/hal_bash_completion
```
lwander: I think you can remove this file as the last command `rm ~/hal_bash_completion`. There is no need to keep it & it'll just clutter the users home directory.

joonlim: Second that.

Looks like you can run ```source <(hal --print-bash-completion)``` and avoid having to create the file altogether.

skim1420 (approved):

PR #744, opened 2018-05-10
Add version edit for Local Git deployment to use HEAD

ttomsu:


setup/install/environment.md
```diff
@@ -123,6 +123,8 @@ the following command.
 
 ```
 hal config deploy edit --type localgit --git-origin-user=<YOUR_GITHUB_USERNAME>
+
+hal config version edit --version branch:upstream/master
```
duftler, line 127: The description just above ends with "...command" but there are two commands.

Seems like this 2nd command should have its own code block with its own single sentence description of why one might want to also issue it after setting the type to `localgit`.

duftler (approved):
LGTM after addressing comment.

dorbin (approved):

setup/install/environment.md
ttomsu: Expanded the text a bit.

duftler (approved):
LGTM.

PR #748, opened 2018-05-11
feat(roadmap): add roadmap

lwander:


community/releases/roadmap.md
```diff
@@ -0,0 +1,121 @@
+---
+layout: single
+title:  "Roadmap"
+sidebar:
+  nav: community
+---
+
```
dorbin, line 7: Might be good to implement zippies, as endabled by PR #713, keep the most recent release expanded, and collapse the rest. Or include a TOC at the top. Or both?

dorbin: Oh, actually now I see that it's not sorted with the latest version at the top. Maybe it should be? How will this be maintained. Will each release be removed from this page once it's released? So we only have, like, two or three releases listed? 

Maybe the thing to do is put them all  in zippies, and let the heading be the release number _and_ the planned date?

lwander: I'm thinking we will remove roadmap entries as releases are published -- will put them in zippies though

lwander:
@dorbin updated by putting each release in a zippy with the feature-freeze date in the header.

dorbin (approved):
Looks good, thanks!

PR #750, opened 2018-05-12
fix(artifacts): fix s3 object example

ethanfrogers:
fixing the example for defining an s3 object artifact. s/name/reference


ethanfrogers:
@lwander @benjaminws PTAL

lwander (approved):
Isn't name also the path?

ethanfrogers:
@lwander good point. i'll add it back. i got thrown off at first because the example didn't work.

ethanfrogers:
@lwander fixed. merge when ready!

PR #754, opened 2018-05-14
Revert "Update jenkins.md to remove `true` from --csrf switch"

dorbin:
Reverts spinnaker/spinnaker.github.io#752


jtk54:
Seems like we only want the `jenkins.md` changes from this PR, right? Seems best to make a new PR with the changes rather than reverting.

dorbin:
That's weird. Will close and re-edit.

PR #756, opened 2018-05-14
docs(setup): Resolved conficts with other PRs, plus further refinements

dorbin:


lwander (approved):

setup/install/providers/index.md
```diff
@@ -1,38 +1,47 @@
...
-* <a href="/setup/install/providers/kubernetes-v2/">Kubernetes V2 (manifest based)</a>
-* <a href="/setup/install/providers/openstack/">Openstack</a>
-* <a href="/setup/install/providers/oracle/">Oracle</a>
-
-See also [`hal config provider`](/reference/halyard/commands/#hal-config-provider).
+Providers](/concepts/providers/) that it supports. So, for Spinnaker to do
+anything you must enable at least one provider, with one Account added for it.
+
+(In fact, if you use a [Distributed
+deployment](/setup/install/environment/#distributed-installation), you'll have a provider (on
```
skim1420: Let's be consistent with the definition of Providers and Accounts above, 3 occasions in this paragraph. Users aren't adding providers, they're adding Accounts.

setup/install/providers/index.md
```diff
@@ -1,38 +1,47 @@
...
+Kubernetes) to run Spinnaker. You can use that one as your only provider, or you
+can create an additional provider for each cloud target.)
```
skim1420: We should prefer not re-using the same account for both running spinnaker and deploying user workloads. I would rather that we prescribe not re-using, then pointing out that they have the option to do so.

setup/install/providers/index.md
```diff
@@ -1,38 +1,47 @@
...
-* <a href="/setup/install/providers/dcos/">DC/OS</a>
-* <a href="/setup/install/providers/gce/">Google Compute Engine</a>
-* <a href="/setup/install/providers/kubernetes/">Kubernetes (legacy)</a>
-* <a href="/setup/install/providers/kubernetes-v2/">Kubernetes V2 (manifest based)</a>
-* <a href="/setup/install/providers/openstack/">Openstack</a>
-* <a href="/setup/install/providers/oracle/">Oracle</a>
-
-See also [`hal config provider`](/reference/halyard/commands/#hal-config-provider).
+Providers](/concepts/providers/) that it supports. So, for Spinnaker to do
+anything you must enable at least one provider, with one Account added for it.
```
skim1420: I prefer the original phrasing of this line. More accurate.

setup/install/providers/index.md
skim1420: Sorry, this comment wasn't addressed. Thanks.

PR #760, opened 2018-05-15
Spells out commands for local development deps

ttomsu:
cc @nvndnl


dorbin (approved):
Is there any value to having 
"Debian: " 
before each command/set of commands. 

If they're all Debian, none of them are Debian. No--that's not what I mean. You know what I mean.

ttomsu:
My intent was to differentiate for different platforms, but can remove that until there is a need for other system instructions

dorbin (approved):
:+1:

PR #767, opened 2018-05-17
docs(spinnaker): add Xiaohongshu to user list

thomassun:
Hi, this is a E-Commerce in China, our homepage is https://www.xiaohongshu.com/ , we are now using spinnaker @kubernetes as the main CD tool,
if we may, We'd like to be listed in https://www.spinnaker.io/community/#whos-using-spinnaker.


community/index.md
```diff
@@ -91,6 +91,9 @@ grubhub_user:
 contaazul_user:
   image_path: assets/images/users/contaazul_logo.svg
   alt: "ContaAzul Logo"
+xiaohongshu_user:
+  image_path: assets/images/users/xiaohongshu_logo.svg
```
skim1420: I believe this should be jpg, not svg.

skim1420: Also, the image file should be placed in assets/images/users/

skim1420 (changes requested):
Welcome to the community! Please fix the mentioned errors, thanks.

community/index.md
dorbin: @skim1420 I think we have both. And svg scales better, right? Do we need to require png?

dorbin: Ah I see. The image you added, @thomassun is a jpg, but you referenced it as an svg. @skim1420 is right, should be jpg.

If you have an svg version, that would be great. Or  a png would also be better than jpg. However jpg is acceptable.

skim1420 (approved):

PR #778, opened 2018-05-18
fix(docs): fix title typo

avram:


dorbin (approved):
Good catches. Thanks, Avram!

PR #779, opened 2018-05-18
docs(readme) Dockerfile that allows page regeneration to be locally tested

pgpx:
Added a Dockerfile that allows you to test changes by running Jekyll within a Docker container instead of installing and running it directly onto your development machine (mounting the repository code as a volume so that changes can still be reflected on the running server without rebuilding the whole image).

Instructions to use (as described in README.md):

```sh
docker build --tag spinnaker/spinnaker.github.io-test .
docker run -it --rm --mount "type=bind,source=$(pwd),target=/code" \
    -p 4000:4000 spinnaker/spinnaker.github.io-test --incremental
```

I'm not sure whether this is something that you would want, but I found it useful when testing some small typo fixes today!  I prefer running such things within Docker if possible.

Note that building the Docker image and running it both log an error to the console, but it seems harmless and I'm not sure how to avoid it:

```
fatal: Not a git repository (or any of the parent directories): .git
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+# Ignore all files to start with, and then override
+*
+!Gemfile*
+!*.gemspec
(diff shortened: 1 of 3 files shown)


lwander (approved):
Nice!

PR #786, opened 2018-05-21
fix(install/minio): disable versioning

lwander:


setup/install/storage/minio.md
```diff
@@ -29,6 +29,14 @@ reachable by Spinnaker. Record the following values:
...
 ## Editing your storage settings
 
+Given that Minio doesn't support versioning objects, we need to disable it
+in Spinnaker. Add the following line to `~/.hal/$DEPLOYMENT/profiles/front50-local.yml`
+(`$DEPLOYMENT` is typically `default`, read more [here](https://www.spinnaker.io/reference/halyard/#deployments)):
+
+```yaml
+spinnaker.s3.versioning: false
+```
+
```
dorbin, line 38: As was already the case before this PR, the line after this one seems to introduce commands, and either there are none, or it's referring to the commands under "Ubuntu installation" and "Docker container installation." I think the latter is the case, but they're h1, so it 's jarring. 
So as long as you're in there, @lwander , could you demote those to h3?

setup/install/storage/minio.md
```diff
@@ -29,6 +29,14 @@ reachable by Spinnaker. Record the following values:
 
 ## Editing your storage settings
 
+Given that Minio doesn't support versioning objects, we need to disable it
+in Spinnaker. Add the following line to `~/.hal/$DEPLOYMENT/profiles/front50-local.yml`
```
dorbin: Put a colon after `front50-local.yml`
and move the sentence about `$DEPLOYMENT` being `default` to _after_ the command.

setup/install/storage/minio.md
lwander: You got it

setup/install/storage/minio.md
```diff
@@ -29,6 +29,14 @@ reachable by Spinnaker. Record the following values:
...
+(`$DEPLOYMENT` is typically `default`, read more [here](https://www.spinnaker.io/reference/halyard/#deployments)):
```
dorbin: Also, s/, read/. Read/

dorbin (approved):
LGTM, pending a few nits.

PR #788, opened 2018-05-22
chore(sizing): Add note about singleton echo and igor.

duftler:


lwander (approved):
Are we sure igor is affected as well?

duftler:
Was mainly reflecting what was written here: https://medium.com/@rizza/re-service-resilience-most-services-can-be-safely-increased-71a8b51b215a

dblackdblack:
thanks for the quick docs update on this.

duftler:
Will wait for Rob to weigh in here as well.

robzienert (approved):

Issue #743, opened 2018-05-10
Outbound links should open new tabs


ttomsu:
Page: [/setup/install/halyard/](https://www.spinnaker.io/setup/install/halyard/)

Feedback:

Going through the setup process, there are a lot of links that go to external pages. I'd rather those pages open up a separate tab so that I can go back to instructions by simply closing tabs.

PR #790, opened 2018-05-22
docs(setup): Changes links to external sites to open in new tabs

dorbin:


dorbin:
#743

dorbin:
@nhayes

ttomsu:
Yay!! :+1:

dorbin:
@ttomsu this is just in /setup/ so far.

setup/artifacts/gcs.md
```diff
@@ -8,15 +8,15 @@ sidebar:
 {% include toc %}
 
 Spinnaker stages that read data from artifacts can consume
-[GCS](https://cloud.google.com/storage/) objects as artifacts.
+[GCS](https://cloud.google.com/storage/){:target="_blank"} objects as artifacts.
```
avh121: Looks like as soon as you add this (on every page) the markdown is in italics until the end of the page, which seems odd to me...do you have any idea why that might be happening? (It just makes me wonder if there's a missing closing brace or something, even though I can't imagine where that would need to go.

setup/features/notifications/index.md
```diff
@@ -115,7 +121,7 @@ Note: your users will need to invite the slack bot to private rooms that want to
 
 ## Twilio
 
-For twilio, you need to add your account [credentials](https://www.twilio.com/help/faq/twilio-basics/what-is-the-auth-token-and-how-can-i-change-it). 
+For twilio, you need to add your account [credentials](https://www.twilio.com/help/faq/twilio-basics/what-is-the-auth-token-and-how-can-i-change-it){:target="_blank"}.
```
avh121: Nit: s/twilio/Twilio, while you're here :)

setup/artifacts/gcs.md
lwander: The `_` character starts italics

ttomsu (approved):

setup/artifacts/gcs.md
dorbin: Escaping it shouldn't work. That is, I would expect it to fix the displayed markdown in editor but break the link in the browser. But it does work.

lwander: I don't think it needs to be escaped... I see you escaped it in the final commit, does it work? I think this is just a quirk of github's syntax highlighting

dorbin: I agree it's not that big a deal if GH (and my editor, btw) highlight as italic, as long as the page renders correctly, but escaping it appears to work fine. I've tested in Chrome, Firefox, and Safari. And  it's better that contributors not have to work in italicized files if it can be avoided.

PR #793, opened 2018-05-23
docs(community): adds links to videos from OpenDev conf 2018

dorbin:


community/presentations.md
```diff
@@ -8,16 +8,22 @@ redirect_from: /docs/presentations
 
 Conference and Meetup Presentations on Spinnaker
 
+### <a href="http://2018.opendevconf.com/" target="_blank">OpenDev CI/CD</a>
+**Videos from OpenDev Conference, Vancouver, BC** \| May 22-23, 2018
+
+* [How Dev/Ops Evolves as the Cloud Does](https://www.youtube.com/watch?v=LjjfWBrvyps){:target="\_blank"}
```
demobox: I think this is "DevOps" (no `/`)? Also, curious about`target="_blank"`? The videos further down the page don't open in a new window, as far as I can make out?

/cc @skim1420

community/presentations.md
```diff
@@ -8,16 +8,22 @@ redirect_from: /docs/presentations
...
+* [Open CD for Open Infrastructure Hybrid and Multi-Cloud Deployments with Spinnaker](https://www.youtube.com/watch?v=EC_zxrk2NQc){:target="\_blank"}
```
demobox: "...Infrastructure: Hybrid..." (add `:`)

demobox (changes requested):
Thanks for adding these, Dave!

community/presentations.md
skim1420: Yeah, omit the slash. slash

dorbin: @demobox , I've been adding the target="_blank", per #743 
It's not non-controversial ;), and I'm open to argument. 
Mean time, I'll add those on this page to be consistent.

PR #801, opened 2018-05-25
docs(pipelines): add doc explaining pipeline management tasks.

avh121:


guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,137 @@
...
+2. Click **Create**, located in the upper right corner of the Pipelines tab.
+  ![](images/create.png)
+3. Choose **Pipeline** from the drop down menu and name your pipeline.
+
+After you create your pipeline, add stages to specify the actions that your
+pipeline will perform.
+
+### Add a stage
+
+Your first stage in any pipeline will be a **Configuration** stage, where you
```
lwander: I don't think this is technically a "stage", even though it sort-of looks like one in the UI.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,137 @@
...
+1. From the pipelines tab, click **Configure** to modify an existing pipeline.
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Delete**.
+  ![](images/pipeline-actions.png)
+
+## Edit as JSON
+
+First, some background: Spinnaker represents pipelines as JSON behind the
+scenes. Any changes you make to your pipeline using the UI are converted to JSON
+when Spinnaker actually runs the pipeline.
```
lwander: I think this should be "when Spinnaker saves the pipeline"

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. From the pipelines tab, click **Configure** to modify an existing pipeline.
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select
+**Edit as JSON**.
+  ![](images/pipeline-actions.png)
+
+## View and restore revision history
+
+Each time you save your pipeline, the current version is added to revision
```
lwander, line 109: I would call out that `minio`, and `redis` do not support revision history.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+history. You can use revision history to diff two versions of a pipeline or to
+restore an older version of a pipeline.
+
+> *Note*: minio and redis do not support version history.
```
jtk54: Maybe can clarify that minio and redis are backing stores for the pipeline repository, and maybe provide a link to another section where you can read more about front50 storage?

jtk54 (approved):
Small comment, LGTM.

guides/user/pipeline/managing-pipelines/index.md
avh121: I updated this to clarify that minio/redis are used to store configuration files, but couldn't find documentation that explains much about front50 storage. Maybe I just missed it though - did you have anything in particular in mind? 

I didn't want to go too in-depth about front50 here since it's only tangentially related, but let me know if you think I should add more detail or mention it more explicitly.

jtk54: It SGTM as is.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
+---
+layout: single
+title:  "Managing Pipelines"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+Pipelines are the essential tool in Spinnaker for configuring how to deploy your
```
dorbin: s/configuring/controlling/  or /determining/.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+application. They are composed of a series of stages that can be combined in
```
dorbin: s/They are/A pipeline is/

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+almost any order, which makes them flexible, consistent, and repeatable.
```
dorbin: s/them/it/

dorbin: Real clear first para!

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+choose **Enable**.
+
+## Delete a pipeline
+
+1. From the pipelines tab, click **Configure** to modify an existing pipeline.
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Delete**.
+  ![](images/pipeline-actions.png)
+
+## Edit as JSON
```
dorbin: "Edit a pipeline as JSON"

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+You can configure your pipelines to run entirely automatically or require manual
+intervention to ensure that everything is working as expected. Similarly,
+pipelines can either be automatically triggered by a wide range of external
+inputs, including other pipelines, or be manually triggered.
+
+This guide explains how to configure and control pipelines, including creating,
+adding triggers to, or disabling pipelines.
+
+## Create a pipeline
+1. Navigate to the **Pipelines** tab in your Spinnaker application.
```
dorbin: Not sure here whether this means "in the Spinnaker UI" or "within an application you've created in Spinnaker." I think the latter, in which case it might be clearer if you have a step before this one saying something like "1. Select the application for which..."

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+pipeline will perform.
+
+### Add a stage
+
+1. Select **Add stage** from your pipeline configuration screen.
+  ![](images/add-stage.png)
+2. Set the stage type using the drop-down menu.
+3. If this is not the first stage in your pipeline, make sure that this stage
+depends on the desired upstream stage(s) using the **Depends on** field.
+  ![](images/stage-depends-on.png)
```
dorbin, line 43: Does the first stage have to be "Configuration"? And is that created automatically? 
(I don't have a running instance at the moment.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+You can add as many stages as your pipeline needs, in any order that makes sense
+for you.
```
dorbin, line 46: If there are any limitations on order, or on what stage types need to preceed others, or anything like that, this might be a good place to mention those, or link to them, for example in the pipeline stages reference. Of course that separate doc doesn't exist yet, but if you don't mention that information here, prepare to link to it once it's doc'd elsewhere.

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Disable**.
+  ![](images/pipeline-actions.png)
+
+In order to re-enable your disabled pipeline, select **Pipeline actions** and
+choose **Enable**.
+
+## Delete a pipeline
+
+1. From the pipelines tab, click **Configure** to modify an existing pipeline.
```
dorbin: s/pipelines/Pipelines/ ?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Delete**.
+  ![](images/pipeline-actions.png)
+
+## Edit as JSON
+
+First, some background: Spinnaker represents pipelines as JSON behind the
```
dorbin: I think you can leave out the "First: some background."

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+When you use the **Edit as JSON** feature, you are directly editing the payload.
+This can be useful for working around limitations of the UI. However, the JSON
+you write here *is not validated* -- **Edit as JSON** essentially allows you to
+modify the pipeline via a free-form textbox. Be careful! It is very easy to
+break the pipeline.
+
+In order to edit your pipeline as JSON:
+
+1. From the pipelines tab, click **Configure** to modify an existing pipeline.
```
dorbin: Pipelines

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+## Edit as JSON
+
+First, some background: Spinnaker represents pipelines as JSON behind the
+scenes. Any changes you make to your pipeline using the UI are converted to JSON
+when Spinnaker saves the pipeline.
+
+When you use the **Edit as JSON** feature, you are directly editing the payload.
+This can be useful for working around limitations of the UI. However, the JSON
+you write here *is not validated* -- **Edit as JSON** essentially allows you to
+modify the pipeline via a free-form textbox. Be careful! It is very easy to
```
dorbin: Might be good to put a warning in between the heading and the intro para.

guides/user/pipeline/managing-pipelines/index.md
dorbin: @jtk54 
@lwander 
If Redis and Minio don't support version history, but we have the versioning functionality, how does that work? It's handled by Front50? If so, it might be worthwhile, @nhayes , to mention something here. But only if it matters to the user. In fact, @lwander , why does the user need to know that Redis and Minio don't support versioning?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,135 @@
...
+  ![](images/revision-history.png)
+2. A button appears that allows you to restore your pipeline to that version. If
+you restore an older version, the current version of your pipeline is saved in
+your revision history in case you want to return to it in the future.
+  ![](images/restore-revision.png)
+
+### Diff two versions of a pipeline
+
+You can compare any version of your pipeline to either the version before it or
+the current pipeline.
```
dorbin, line 140: @lwander is this the correct behavior?

dorbin (changes requested):

guides/user/pipeline/managing-pipelines/index.md
avh121: Yes and yes; I removed the note I had about that because @lwander pointed out that "Configuration" is not technically a stage. I don't think that that technicality is worth noting here, though, and I'm open to adding the note about "Configuration" being the first step of every pipeline back in here.

guides/user/pipeline/managing-pipelines/index.md
avh121: As far as I know there aren't specific limitations on order along the lines of "stage X must come before stage Y", although there are stages that might not be successful on their own (e.g., you need an image to deploy before you can deploy it, a Server Group has to exist before you can disable it, etc). I think cases like the examples I just gave are specific and common sense enough that they would be hard to document here in a useful way, though.

@lwander or @jtk54 do you know if I'm correct about the lack of explicit ordering limitations? If not, do you know what they are?

guides/user/pipeline/managing-pipelines/index.md
dorbin: Yeah, I think you can mention that it's automatically created and is the first "step" (is a good word for it), and offhandedly mention the technicality that it's not a stage. Up to you tho.

avh121: Added this back in without directly explaining that Configuration isn't a stage -- let me know what you think.

dorbin: :+1:

dorbin (approved):

guides/user/pipeline/managing-pipelines/index.md
lwander: Yes that sounds right

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+  ![](images/add-stage.png)
+2. Set the stage type using the drop-down menu.
+3. If this is not the first stage in your pipeline, make sure that this stage
+depends on the desired upstream stage(s) using the **Depends on** field.
+  ![](images/stage-depends-on.png)
+
+You can add as many stages as your pipeline needs, in any order that makes sense
+for you.
+
+### Add a trigger
```
demobox, line 48: Curious about having this after vs. before "Create a stage"?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+1. From the **Pipelines** tab, click **Configure** to modify an existing
+pipeline.
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Disable**.
+  ![](images/pipeline-actions.png)
+
+In order to re-enable your disabled pipeline, select **Pipeline actions** and
+choose **Enable**.
+
+## Delete a pipeline
```
demobox, line 77: If we are not documenting all the actions (I think we're not documenting "rename", for example), is there a specific reason for including this? "Delete" seems fairly self-explanatory, and since this section has screenshots that may need to be updated it adds to our potential maintenance load?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+pipeline.
+  ![](images/select-configure.png)
+2. Click **Pipeline actions** in the upper right corner, and select **Delete**.
+  ![](images/pipeline-actions.png)
+
+## Edit a pipeline as JSON
+
+> **Warning**: the JSON you write here is not validated -- Edit as JSON
+essentially allows you to modify the pipeline via a free-form textbox. **Be
+careful!** It is very easy to break the pipeline.
```
demobox, line 89: Is there any tip we can add here? Such as "you can revert to the previous version if you make a mistake" or so?

/cc @lwander

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+Spinnaker represents pipelines as JSON behind the scenes. Any changes you make
+to your pipeline using the UI are converted to JSON when Spinnaker saves the
+pipeline.
+
+When you use the **Edit as JSON** feature, you are directly editing the payload.
+This can be useful for working around limitations of the UI.
```
demobox, line 96: This isn't very confidence-inspiring :-( Do we have any particular examples of that?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+**Edit as JSON**.
+  ![](images/pipeline-actions.png)
+
+## View and restore revision history
+
+Each time you save your pipeline, the current version is added to revision
+history. You can use revision history to diff two versions of a pipeline or to
+restore an older version of a pipeline.
+
+> *Note*: If you are using minio or redis to store your Spinnaker configuration
```
demobox, line 113: How would a user or admin know if this is the case? Is this about having chosen minio or redis as your [storage service](https://www.spinnaker.io/setup/install/storage/)? If so, perhaps rephrase to use that wording?

guides/user/pipeline/managing-pipelines/index.md
```diff
@@ -0,0 +1,146 @@
...
+### Restore a previous version
+
+1. Choose a version from the **Revision** drop-down.
+  ![](images/revision-history.png)
+2. A button appears that allows you to restore your pipeline to that version. If
+you restore an older version, the current version of your pipeline is saved in
+your revision history in case you want to return to it in the future.
+  ![](images/restore-revision.png)
+
+### Diff two versions of a pipeline
```
demobox, line 137: This seems like an option you'd try _before_ restoring to a previous version?

guides/user/pipeline/managing-pipelines/index.md
avh121: I'm actually not sure - @lwander, @jtk54 could you clarify?

lwander: Yes it would have to be chosen explicitly during setup/configuration

guides/user/pipeline/managing-pipelines/index.md
lwander: Maybe can be worded as "This can be useful to set fields/properties not exposed by the UI"

avh121: Fair point. There are some stages or pipeline features that aren't surfaced in the UI. For example I think it's technically possible to have a pipeline create a load balancer for you, but there's no UI stage for that. I think there are other examples too, but I haven't dug into it too deeply.

The reason why I didn't go into this more thoroughly is that Edit as JSON is a pretty complex/advanced thing to do, and really deserves its own separate user guide. I wanted to give a high level explanation of it here, but I don't think this is the place to go into a lot of detail about specifically what you can do with it.

guides/user/pipeline/managing-pipelines/index.md
lwander: Yeah you can role back if you have versioning enabled like explained below. Maybe we should make a callout here?

guides/user/pipeline/managing-pipelines/index.md
avh121: I mostly included it because I figured that people might reasonably google 'delete spinnaker pipeline' and I wanted it to show up. FWIW, it reuses screenshots that are also used in explaining other actions, so it doesn't actually add to the maintenance burden.

guides/user/pipeline/managing-pipelines/index.md
avh121: Yep, good point (here and above about stages/triggers). I'll open another PR to reorder this and address some of your other comments.

guides/user/pipeline/managing-pipelines/index.md
demobox: > because I figured that people might reasonably google 'delete spinnaker pipeline' and I wanted it to show up

Ah, clear; thanks for explaining!

PR #808, opened 2018-05-30
docs(halyard-gke-deploy-rbac): Fix base64 option

Mahito:
`base64 -D` didn't work well as follow.

```
$ kubectl get secret $SERVICE_ACCOUNT_TOKEN -o jsonpath='{.data.token}' | base64 -D
base64: invalid option -- 'D'
Try 'base64 --help' for more information.
```

option `-D` worked on Mac, but it did not work on Linux because it is` -d` on Linux.
This PR has changed to option `--decode` which works on both MacOS and Linux.


lwander (approved):
Thanks!

PR #813, opened 2018-05-30
docs(permissions): Add permissions doc for authorization.

stewchen:
This should not go in until halyard releases with permissions support.

### DIFF
--- a/setup/security/authorization/index.md
+++ b/setup/security/authorization/index.md
@@ -68,17 +68,31 @@ accounts.
 ### Accounts
 In the dark ages (before Fiat), only accounts could be restricted. Because
 Clouddriver is the source of truth for accounts, Fiat reaches out to Clouddriver
-to gather the list of available accounts. To add access restrictions to an
-account, update your `halconfig` to include the `requiredGroupMembership` field.
+to gather the list of available accounts. 
+
+There are two types of access restrictions to an account, `READ` and `WRITE`. Users must have
+at least one `READ` permission of an account to view the account's cloud resources, and at least one
+`WRITE` permission to make changes to the resources.
+
+These halyard commands manage the `READ` and `WRITE` permissions.
 
 ```bash
 PROVIDER= # Your cloud provider
-GROUP=    # The new group membership
 
-hal config provider $PROVIDER account edit $ACCOUNT --add-required-group-membership $GROUP
+hal config provider $PROVIDER account edit $ACCOUNT \
+  --add-read-permission role1 \ # Adds a READ permission
+  --add-write-permission role2 \ # Adds a WRITE permission
+  --remove-read-permission role3 \ # Removes a READ permission
+  --remove-write-permission role4 # Removes a WRITE permission
+
+# Alternatively, you can overwrite the whole read or write list, comma delimited.
+hal config provider $PROVIDER account edit $ACCOUNT \
+  --read-permissions role1,role2,role3 \
+  --write-permissions role1,role2,role3
 ```
 
-Alternatively, you can overwrite the whole list using the `--required-group-membership` flag.
+(Deprecated) `requiredGroupMembership` is the old way to add access restrictions to an account. 
+This method does not distinguish between `READ` and `WRITE` - users with access will have both.
 
 
 ### Applications


setup/security/authorization/index.md
```diff
@@ -68,17 +68,41 @@ accounts.
...
+
+# Alternatively, you can overwrite the whole read or write list, comma delimited.
+hal config provider $PROVIDER account edit $ACCOUNT \
+  --read-permissions role1,role2,role3 \
+  --write-permissions role1,role2,role3
 ```
 
-Alternatively, you can overwrite the whole list using the `--required-group-membership` flag.
+(Deprecated) `requiredGroupMembership` is the old way to add access restrictions to an account. 
+This method does not distinguish between `READ` and `WRITE`.
```
dorbin: @stewchen 
Just out of curiosity, what was the effect when we didn't distinugish between read and write? User would have both?

setup/security/authorization/index.md
```diff
@@ -68,17 +68,41 @@ accounts.
 ### Accounts
 In the dark ages (before Fiat), only accounts could be restricted. Because
 Clouddriver is the source of truth for accounts, Fiat reaches out to Clouddriver
-to gather the list of available accounts. To add access restrictions to an
-account, update your `halconfig` to include the `requiredGroupMembership` field.
+to gather the list of available accounts. 
+
+There are two types of access restrictions to an account, `READ` and `WRITE`. Users must have
+matching roles to all the `READ` permissions of an account to view the account's cloud resources.
```
dorbin: @stewchen 
Would it be correct to say "User must have `READ` access to an account to view..." rather than "..must have matching roles to all the `READ` permissions..."
(What is the importance of the "matching roles" part?)

dorbin (approved):
A couple of questions that are mostly curiosity, but LGTM.

setup/security/authorization/index.md
stewchen: Yup

setup/security/authorization/index.md
stewchen: I was trying to imply that a user would need to have all `READ` roles if there were multiple, to be able to read. I'll just separate these into two sentences hah.

setup/security/authorization/index.md
lwander: Do we know at which version of spinnaker this stopped being accepted? Maybe we can call that out here

stewchen: Actually, I misspoke - they are still functional - but I guess people are unhappy with the deprecation messages introduced in https://github.com/spinnaker/clouddriver/pull/1642. How can I find the clouddriver version and/or spinnaker release that introduced that commit?

setup/security/authorization/index.md
```diff
@@ -68,17 +68,42 @@ accounts.
...
 In the dark ages (before Fiat), only accounts could be restricted. Because
 Clouddriver is the source of truth for accounts, Fiat reaches out to Clouddriver
-to gather the list of available accounts. To add access restrictions to an
-account, update your `halconfig` to include the `requiredGroupMembership` field.
+to gather the list of available accounts. 
+
+There are two types of access restrictions to an account, `READ` and `WRITE`. Users must have
+`READ` permissions of an account to view the account's cloud resources, and`WRITE` permissions to
+make changes to the resources. If there are multiple permissions for one type of access on an
+account, users must have matching roles to all of them to be granted that type of access.
```
ttomsu: > If there are multiple permissions for one type of access on an account, users must have matching roles to all of them to be granted that type of access.

This statement is unclear to me. A user must have at least one of the listed roles in order to be granted access, but it sounds like this statement is saying one needs to have _all_ of the listed roles.

setup/security/authorization/index.md
ttomsu: In the commit, it lists what tags this change was included with:

yfftd8sq4q3

It appears this code has been included since the 1.0.0 release. Hooray!

setup/security/authorization/index.md
```diff
@@ -68,17 +68,42 @@ accounts.
...
+hal config provider $PROVIDER account edit $ACCOUNT \
+  --read-permissions role1,role2,role3 \
+  --write-permissions role1,role2,role3
 ```
 
-Alternatively, you can overwrite the whole list using the `--required-group-membership` flag.
+(Deprecated) `requiredGroupMembership` is the old way to add access restrictions to an account. 
+This method does not distinguish between `READ` and `WRITE`.
+
+```bash
```
ttomsu: I would remove the examples of the deprecated way.

ttomsu (approved):

PR #814, opened 2018-05-31
fix(halyard/custom): s/Lifecycle/LifeCycle/

lwander:
https://github.com/spinnaker/spinnaker/issues/2851

### DIFF
--- a/reference/halyard/custom.md
+++ b/reference/halyard/custom.md
@@ -87,7 +87,7 @@ cluster.
 | `port` | The port number this service is bound to, and will accept requests on. |
 | `safeToUpdate` | Whether or not this service can be shutdown, and spun on a new VM/container. This protects datastores like Vault & Redis from being taken down from Halyard. |
 | `scheme` | The URI scheme used to address this service, e.g. `http` vs.  `https`. |
-| `skipLifecycleManagement` | Whether or not Halyard should skip managing a service's lifecycle. |
+| `skipLifeCycleManagement` | Whether or not Halyard should skip managing a service's life cycle. |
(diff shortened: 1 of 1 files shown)


reference/halyard/custom.md
```diff
@@ -87,7 +87,7 @@ cluster.
 | `port` | The port number this service is bound to, and will accept requests on. |
 | `safeToUpdate` | Whether or not this service can be shutdown, and spun on a new VM/container. This protects datastores like Vault & Redis from being taken down from Halyard. |
 | `scheme` | The URI scheme used to address this service, e.g. `http` vs.  `https`. |
-| `skipLifecycleManagement` | Whether or not Halyard should skip managing a service's lifecycle. |
+| `skipLifeCycleManagement` | Whether or not Halyard should skip managing a service's life cycle. |
 | `targetSize` | The initial number of nodes this service will be created with.  This is only respected on the initial deployment, and further edits will be rejected in favor of the prior service version's size. |
```
dorbin, line 91: Actually, [one word](https://developers.google.com/style/word-list#letter-L).

lwander: Darn....

If we put it back into one word, is it jarring that the flag has the C capitalized?

dorbin: Not to me. 
IMO, we should do whatever has the least negative impact on the user. If [things work better](https://github.com/spinnaker/spinnaker/issues/2851) when it's skipLifeCycleManagement, I'm ok with violating the style in doc text.

lwander: Ok, let's keep it as-is for now

PR #815, opened 2018-05-31
docs(guides): added config for using kayenta with Minio

e96wic:


setup/canary/index.md
```diff
@@ -350,3 +350,33 @@ hal config canary prometheus account get
...
+
+## Set up Minio as storage provider
+Configure your canary analysis to use Prometheus as your telemetry provider.
+
+### Add a custom halyard profile for kayenta
+```
+  aws:
+    enabled: true
+    accounts:
+    - name: <name>
```
lwander, line 365: Is this only possible using a profile, not using the halyard commands? I'm all for showing both, but for consistency with the rest of the doc we should probably show the hal commands. 

@duftler

e96wic: Right now there is no option in halyard to specify the endpoint which is important for this to work. You can specify this for the front50 config but not yet for the kayenta config.

duftler: Ah, hadn't realized that. I'll have a look and add it the necessary option to halyard. Let's leave this open for the moment while I dig into it.

duftler:
Adds support for `endpoint`: https://github.com/spinnaker/halyard/pull/968

cfieber:
@duftler is this mergeable?

duftler:
With the endpoint support in halyard, I don't think this is still necessary. Closing.

PR #816, opened 2018-05-31
docs(reference): add reference doc for pipeline stages.

avh121:
Note that the intention of this PR is just to create a doc listing the existing stages and providing a brief explanation for each one. In the future I'll go through and flesh out the descriptions to make this into a more fully-fledged reference guide.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -365,6 +365,8 @@ reference:
         url: /reference/architecture/load/
       - title: Life of a Bake
         url: /reference/architecture/loab/
+  - title: Pipeline Stages
+    url: /reference/pipeline/stages
 
 community:
   - title: Releases
--- a/reference/pipeline/stages/index.md
+++ b/reference/pipeline/stages/index.md
@@ -16,77 +16,75 @@ Spinnaker pipelines.
 ## General
 
 ### Bake
-Bakes an image in the specified region.
+Bake an image in the specified region.
 
 ### Canary Analysis
-Runs automated canary analysis against the deployment before fully deploying.
-For more information on automated canary analysis in Spinnaker, see the
-[user guide](/guides/user/canary/).
+Run [automated canary analysis](/guides/user/canary/) against the deployment
+before fully deploying.
 
 ### Check Preconditions
-Checks for preconditions before continuing. Specifically, you can either check
-that your cluster is a particular size, or add a pipeline expression. See the
+Check for preconditions before continuing. For example, you can check that
+your cluster is a particular size, or add a pipeline expression. See the
 [pipeline expressions guide](/guides/user/pipeline-expressions/) for more
 information about creating and using pipeline expressions.
 
 ### Clone Server Group
-Deploys a new Server Group which is a copy of the specified Server Group.
+Deploy a new Server Group that is a copy of the specified Server Group.
 
 ### Deploy
-Deploys the previously baked or found image using the specified deployment
+Deploy the previously baked or found image using the specified deployment
 strategy. Spinnaker provides built-in support for both red/black (also known as
-blue/green) and highlander deployment strategies. You can also choose to deploy
+blue/green) and Highlander deployment strategies. You can also choose to deploy
 with no impact on existing Server Groups, or build your own custom deployment
 strategy.
 
 ### Destroy Server Group
-Deletes a Server Group and its resources from the specified Cluster. You must
+Delete a Server Group and its resources from the specified Cluster. You must
 specify whether you want to delete the newest, oldest, or previous
 (second-most-recently deployed) Server Group when this stage starts.
 
 ### Disable Cluster
-Disables the specified Cluster, which means that the cluster remains up but
+Disable the specified Cluster, which means that the cluster remains up but
 stops handling traffic. If desired, you can leave a specific number of Server
 Groups running while the rest of the cluster is disabled.
 
 ### Disable Server Group
-Disables a Server Group, which means that the Server Group remains up but stops
+Disable a Server Group, which means that the Server Group remains up but stops
 handling any traffic. This makes it easy to both route traffic to a new
 Server Group and roll back those changes if necessary. You must choose whether
 to disable the Server Group which is newest, oldest, or previous
 (second-most-recently deployed) when this stage starts.
 
 ### Enable Server Group
-Enables a Server Group, which means that the Server Group will begin handling
-traffic again.
+Enable a Server Group, which means that the Server Group will will resume
+handling traffic.
 
 ### Find Artifact From Execution
 Find and bind an artifact from another pipeline execution.
 
 ### Find Image From Cluster
-Finds an image to deploy from an existing Cluster. You must specify the Cluster,
-Server Group, and image name such that there is exactly one match in order for
-this stage to succeed.
+Find an image to deploy from an existing Cluster. You must specify the Cluster,
+Server Group, and image name such that there is exactly one match.
 
 ### Find Image From Tags
-Finds an image to deploy from tags.
+Find an image to deploy from tags.
 
 ### Jenkins
-Runs a Jenkins job. You must [set up Jenkins](/setup/ci/jenkins/) in order to
+Run a Jenkins job. You must [set up Jenkins](/setup/ci/jenkins/) in order to
 use this stage.
 
 ### Manual Judgment
-Waits for user approval before continuing.
+Wait for the user to click **Continue** before continuing.
 
 ### Pipeline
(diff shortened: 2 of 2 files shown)


_data/navigation.yml
```diff
@@ -347,6 +347,8 @@ reference:
         url: /reference/architecture/load/
       - title: Life of a Bake
         url: /reference/architecture/loab/
+  - title: Pipeline Stages
```
lwander, line 368: Should we maybe put this under a collapsible Pipelines header and call it Stages? I'm sure we can add more pipeline reference material for that header.

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,122 @@
...
+layout: single
+title:  "Pipeline Stages"
+sidebar:
+  nav: reference
+---
+
+{% include toc %}
+
+This article describes the currently-supported stages that you can add to your
+Spinnaker pipelines.
```
lwander: Not sure what @dorbin 's stance on this is, but given that this is intentionally a bare-bones first pass, is it worth calling that out here?

lwander (approved):

_data/navigation.yml
dorbin: Such as what? Do we need reference material for triggers?

reference/pipeline/stages/index.md
dorbin: Not sure what you mean, @lwander . Are you referring to the "currently-supported" part?

_data/navigation.yml
dorbin: If so, then 👍

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,122 @@
...
+
+## General
+
+### Bake
+Bakes an image in the specified region.
+
+### Canary Analysis
+Runs a canary task.
```
dorbin: Runs automated canary analysis against the deployment before fully deploying.

For this one, and probably for others, you can link to more information. In this case /guides/user/canary/

dorbin (changes requested):
What I wrote, plus what we talked about.

_data/navigation.yml
avh121: I decided to do it this way since we don't have other pipeline reference content at the moment, but agreed that as soon as we add more we should create a Pipelines dropdown. Also open to doing it now if you prefer, but my slight preference is to wait until it's actually necessary.

lwander: I guess it won't require adding more redirects so if we want to defer until more content comes that sounds good to me. 

Triggers, deep dives into stages vs. tasks, synthetic stages, timeouts, SpEL etc... could all be useful to explore.

reference/pipeline/stages/index.md
lwander: I would have recommended calling out that this is just a first pass, and that we will be adding details on each stage shortly. Right now each description is a rewording of the stage title and makes the doc look a little out-of-place in a reference section.

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+## General
+
+### Bake
+Bakes an image in the specified region.
+
+### Canary Analysis
+Runs automated canary analysis against the deployment before fully deploying.
+For more information on automated canary analysis in Spinnaker, see the
+[user guide](/guides/user/canary/).
```
dorbin: I think you could delete this whole 2nd sentence and just link "automated canary analysis" in the first sentence, to this target.

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Check Preconditions
+Checks for preconditions before continuing. Specifically, you can either check
+that your cluster is a particular size, or add a pipeline expression. See the
```
dorbin: If these are the only two examples, and will probably remain so, then "Specifically..." is ok. But otherwise I would change this to "For example, you can..."

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+[pipeline expressions guide](/guides/user/pipeline-expressions/) for more
+information about creating and using pipeline expressions.
+
+### Clone Server Group
+Deploys a new Server Group which is a copy of the specified Server Group.
```
dorbin: s/which/that/

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Deploy
+Deploys the previously baked or found image using the specified deployment
+strategy. Spinnaker provides built-in support for both red/black (also known as
+blue/green) and highlander deployment strategies. You can also choose to deploy
```
dorbin: s/highlander/Highlander/

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+### Disable Server Group
+Disables a Server Group, which means that the Server Group remains up but stops
+handling any traffic. This makes it easy to both route traffic to a new
+Server Group and roll back those changes if necessary. You must choose whether
+to disable the Server Group which is newest, oldest, or previous
+(second-most-recently deployed) when this stage starts.
+
+### Enable Server Group
+Enables a Server Group, which means that the Server Group will begin handling
+traffic again.
```
dorbin: "again" implies that it was enabled, then disabled, and this is now re-enabling. But would this stage ever be used to enable a server group *originally*?  I'm guessing no, but I just don't know.

dorbin: Oh, and if this is right--that it's *re*-enabling--you could shorten it:
s/will begin handling traffic again/will resume handling traffic/

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Find Artifact From Execution
+Find and bind an artifact from another pipeline execution.
+
+### Find Image From Cluster
+Finds an image to deploy from an existing Cluster. You must specify the Cluster,
+Server Group, and image name such that there is exactly one match in order for
+this stage to succeed.
```
dorbin: I would either delete "in order for this stage to succeed." or move it to the beginning of the setence. I lean slightly in favor of deleting, because "You must..."

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Find Image From Tags
+Finds an image to deploy from tags.
+
+### Jenkins
+Runs a Jenkins job. You must [set up Jenkins](/setup/ci/jenkins/) in order to
+use this stage.
+
+### Manual Judgment
+Waits for user approval before continuing.
```
dorbin: Waits for the user to click [whatever the button name is] before continuing.

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Pipeline
+Runs an existing pipeline: you can select any pipeline from this application and
+run it as a sub-pipeline.
+
+
+### Resize Server Group
+Resize the oldest, newest, or previous (second-most-recently deployed) Server
```
dorbin: "...oldest, newest, or second newest..."

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+Group. You can resize the Server Group by either a percentage of its current
+size or a specific amount. The resizing strategies available to you are:
```
dorbin: I think you can delete "available to you" or at least "to you".

reference/pipeline/stages/index.md
```diff
@@ -0,0 +1,161 @@
...
+to match the largest Server Group in the Cluster. Optionally, you can specify
+additional capacity to add as well.
+* **Scale to Exact Size**, which adjusts the size of the target Server Group to
+match the specified capacity.
+
+### Rollback Cluster
+Roll back one or more regions in a Cluster.
+
+### Run Job
+Runs a container.
```
dorbin: Some of these first-word verbs are indicative and some are imperative. "Runs," here, versus "Roll back" in the previous one and "Scale" in the next one. Kind of a nit, but it might be cleaner to be consistent throughout.

dorbin (changes requested):

reference/pipeline/stages/index.md
avh121: Yep, double checked and it seems like it's always re-enabling - will update this to use  your wording :)

dorbin (approved):
Looks great, Nora, thanks!

PR #817, opened 2018-05-31
feat(scaling/caching): notes on caching configuration

lwander:
Fixed a typo in the externalize redis docs as well

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -161,6 +161,8 @@ setup:
       - title: Configure Caching
         collapsed: true
         children:
+          - title: Configure Redis Usage
+            url: /setup/productionize/caching/configure-redis-usage/
           - title: Externalize Redis
             url: /setup/productionize/caching/externalize-redis/
 
@@ -207,8 +209,6 @@ guides:
       - title: Pipelines
         collapsed: true
         children:
-          - title: Managing Pipelines
-            url: /guides/user/pipeline/managing-pipelines/
           - title: Triggering Pipelines
             collapsed: true
             children:
--- a/setup/productionize/caching/configure-redis-usage.md
+++ b/setup/productionize/caching/configure-redis-usage.md
@@ -1,0 +1,68 @@
+---
+layout: single
+title:  "Configure Spinnaker's Usage of Redis"
+sidebar:
+  nav: setup
+---
+
+{% include toc %}
+
+Spinnaker relies on its Redis cache for a number of reasons: caching
+infrastructure, storing live executions, returning pipeline definitions faster,
+etc... Most of these can be configured to your needs, whether it is to make
+Spinnaker more responsive, or reduce the load on a downstream dependency.
+
+## Configure infrastructure caching
+
+By default, Spinnaker's Clouddriver service spawns "caching agents" every
+30 seconds. These agents are scheduled and coordinated across all instances of
+Clouddriver, allowing you to reduce the load on any single instance of
+Clouddriver by creating more replicas of the service.
+
+To adjust how this caching happens, Clouddriver exposes a few properties that
+can be set in `~/.hal/$DEPLOYMENT/profiles/clouddriver-local.yml`:
+
+```yaml
+# How many seconds (default 30) between runs of agent. Lowering this number
+# means the resources in the Spinnaker UI will be updated more frequently,
+# at the cost higher API/quota usage of your cloud provider.
+redis.poll.intervalSeconds:
+
+
+# How many seconds (default 600, 5 minutes) Clouddriver will wait to reschedule
+# an agent that never completes (never throws an error or returns cache data).
+# If your agents are taking a long time to complete their cache cycles
+# successfully and Clouddriver is prematurely rescheduling them, you can try to
+# raise this number.
+redis.poll.timeoutSeconds:
+```
+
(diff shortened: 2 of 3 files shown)


setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
+---
+layout: single
+title:  "Configure Redis Usage"
```
dorbin: Instead of "Usage" maybe something more specific, like "Configure Redis to scale up Spinnaker" or something more exact about what you're configuring Redis for?

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+sidebar:
+  nav: setup
+---
+
+{% include toc %}
+
+Spinnaker relies on its Redis cache for a number of reasons: caching
+infrastructure, storing live executions, returning pipeline definitions faster,
+etc... Each of these can be configured to your needs, whether it is to make
+Spinnaker more responsive, or reduce the load on a downstream dependency.
```
dorbin: The list of things we use Redis cache for does not map to the sections below.  That might be ok, but because you say "Each of these can be configured" it sounds like you're not delivering on a promise. 

It might be good to keep this list, all (or a representative list of) the things that we rely on Redis for, then instead of saying "Each of these...," say "You can configure Redis to optimizes some of these for scale..."

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Configure infrastructure caching
+
+By default, Spinnaker's Clouddriver service will spawn "caching agents" every
```
dorbin: s/will spawn/spawns/

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+Clouddriver, allowing you to reduce the load on any single instance of
+Clouddriver by creating more replicas of the service.
+
+To adjust how this caching happens, Clouddriver exposes a few properties that
+can be set in `~/.hal/$DEPLOYMENT/profiles/clouddriver-local.yml`:
+
+```yaml
+# How many seconds (default 30) between runs of agent. Lowering this number
+# means the resources in the Spinnaker UI will be updated more frequently,
+# at the cost higher API/quota usage of your cloud provider.
```
dorbin: So, to optimize for scale use a higher number?

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you see that your Redis memory usage is growing unbounded, it's most likely
+due to the fact that the orchestration engine stores all pipeline executions
+forever by default. This can be configured by setting the following properties
+in `~/.hal/$DEPLOYMENT/profiles/orca-local.yml`:
+
+```yaml
+pollers:
+  oldPipelineCleanup:
+    enabled: true       # This enables old pipline cleanup.
```
dorbin: s/pipline/pipeline/

Are the comments in these yaml files written by you, on this occasion, or were they already there?

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+redis.poll.timeoutSeconds:
+```
+
+> `$DEPLOYMENT` is typically `default`. See [the
+> documentation](/reference/halyard#deployments) for more details.
+
+## Configure pipeline cleanup
+
+If you see that your Redis memory usage is growing unbounded, it's most likely
+due to the fact that the orchestration engine stores all pipeline executions
```
dorbin: s/most likely due to the fact that/probably because/

setup/scaling/caching/configure-redis-usage.md
```diff
@@ -0,0 +1,68 @@
...
+forever by default. This can be configured by setting the following properties
+in `~/.hal/$DEPLOYMENT/profiles/orca-local.yml`:
```
dorbin: s/configured by setting the following properties/changed/

dorbin (changes requested):
Because this is all in the context of scaling, you might add comments in the yaml files, for each property, which way to turn the dial to scale up.

setup/scaling/caching/configure-redis-usage.md
lwander: Possibly -- unless you need fresher data.

setup/scaling/caching/configure-redis-usage.md
lwander: These yaml files exist here only

lwander:
It's not that clear. The purpose as described in the scaling header is both provide intuition as well as some guidance/tips. Most of the explanations here fall in the former. 

If it were just as easy as "make these numbers bigger/smaller" then we wouldn't have to explain any of this to the user, we could just do it for them via halyard.

setup/scaling/caching/configure-redis-usage.md
lwander: See my comment below -- it's not that clear cut. We could say "Configure Spinnaker's usage of Redis" maybe?

dorbin (approved):
taking into account the conversation we just had, and with other edits...LGTM

PR #825, opened 2018-06-04
docs(setup): Adds Google I/O Codelab to quickstart page

skim1420:
@demobox, @lwander FYI, any objections/thoughts to this?


dorbin (commented):
I was thinking of adding it to /guides/tutorials/codelabs/. Can we do both?

skim1420:
sgtm, or if you're inclined to, even remove this one and only add to codelabs. Just wanted it represented somewhere, and I thought quickstart was a decent one.

On a broader sense, it raises what the purpose of the quickstart section is vs codelabs. We had discussed the notion that a general purpose quickstart often sets up wrong expectations with the user.

demobox:
> even remove this one and only add to codelabs

I think adding it here risks confusing users in that the resulting instance here isn't accessible in the same way as one created by the Helm chart or Cloud Launcher, e.g. it's not created in a user's project or account.

+1 adding it [guides](https://www.spinnaker.io/guides/) and/or codelabs.

skim1420:
sgtm, closing.

PR #827, opened 2018-06-04
feat(reference/kubernetes): max-version-history docs

lwander:


lwander:
https://github.com/spinnaker/clouddriver/pull/2688

reference/providers/kubernetes-v2/index.md
```diff
@@ -130,6 +130,20 @@ command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
   ConfigMap or Secret to be deployed without appending a new version when the
   contents change, for example.
 
+* `strategy.spinnaker.io/max-version-history`
+
+  When set to a non-negative integer, this configures how many versions of a
+  resource to keep around. When more than `max-version-history` versions of a
+  Kubernetes artifact exist, Spinnaker deletes all older versions.
```
dorbin, line 137: Doesn't just delete the oldest each time? From 1 version to max number of them, then back to 1 each time?

reference/providers/kubernetes-v2/index.md
```diff
@@ -130,6 +130,20 @@ command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
...
+  __Resources are sorted by the `metadata.creationTimestamp` kubernetes property
+  rather than the version number.__
+
+  Keep in mind, if you are trying to restrict how many copies of a ReplicaSet
+  a Deployment is managing, that is configured by
+  [`spec.revisionHistoryLimit`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy).
+  If instead Spinnaker is deploying ReplicaSets directly without a Deployment,
+  this annotation does the job.
```
dorbin, line 145: I wonder if there's a way to get this information at the top of the section, the use case for this property contrasted with the other property, before the "When set to a non-negative integer..." information.

dorbin (approved):
Just a couple of questions, but lgtm

reference/providers/kubernetes-v2/index.md
lwander: Usually it does, but if you either go from not having this property set to setting it, or shrink the number, more than 1 will be deleted.

reference/providers/kubernetes-v2/index.md
lwander: I added this to clear up a potential confusion, I don't think this in particular should go at the top. In general I feel with the reference docs we can be a little more abstract. Maybe we should have a guide for this?

dorbin: Cool by me. At first glance I don't see an obvious place in your existing k8sv2 guide for this. Feel free to follow up with me about what might be written there, but I'm ok with leaving this here.

reference/providers/kubernetes-v2/index.md
dorbin: :+1:

reference/providers/kubernetes-v2/index.md
lwander: I've assigned myself to a new guide to handle this: https://github.com/spinnaker/spinnaker.github.io/issues/829

I imagine it'll be pretty short & look similar to one of these: https://www.spinnaker.io/guides/user/kubernetes-v2/deploy-manifest/

PR #833, opened 2018-06-06
fix(kubernetes): update service account docs

ethanfrogers:
some of the references didn't match and you need to base64 decode the
token from the secret

@lwander PTAL


lwander (approved):
Good catch, thanks!

PR #837, opened 2018-06-07
docs(providers/kubernetes): Fix typo

frelon:


dorbin (approved):
Good catch, Fredrik. Thanks!

PR #838, opened 2018-06-07
Fix incosistent numbering in the install proceedure

dj80hd:
On the [install page](https://www.spinnaker.io/setup/install/), "Choose an environment" is 2.
However, [the environment page](https://www.spinnaker.io/setup/install/environment/) shows it as 3.
Likewise, "Choose Cloud Providers" is 3 on the [install page](https://www.spinnaker.io/setup/install/), but [the providers page](https://www.spinnaker.io/setup/install/providers/) shows 2.

This PR makes the numbering consistent.

### DIFF
--- a/setup/install/index.md
+++ b/setup/install/index.md
@@ -14,8 +14,8 @@ in [Quickstart](/setup/quickstart/) might be a better choice.
 
 Installing a complete Spinnaker involves these steps:
 1. [Install Halyard](/setup/install/halyard/)
-2. [Choose an environment](/setup/install/environment/)
-3. [Choose a cloud provider](/setup/install/providers/)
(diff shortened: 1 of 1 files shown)


dorbin (changes requested):
Hi, @dj80hd 
Good catch!

But the place to change it is on the install page (/setup/install), to match the nav and titles. We
need providers to come first because the setup for environment requires the  $ACCOUNT that's set when configuring the provider. 

You want to do that, or would you like me to?

dj80hd:
ty @dorbin ; should be ok now.

dorbin (approved):
:+1: 
Thanks, @dj80hd

PR #839, opened 2018-06-08
fix(docs): instructions for email

tomaslin:
fix example for email using hotmail as an example since gmail has tighten down on security substantially and the previous instructions don't work anymore.


lwander (approved):

dorbin (commented):
Just out of curiousity, Tomás, will we be able to make this work with gmail at some point?

dorbin (approved):

tomaslin:
@dorbin I wasn't able to get it to work with gmail due to inability to generate application tokens. It's really more of a question for Google as it seems they want to discourage the use of gmail smtp to send emails...

PR #840, opened 2018-06-08
docs(reference): more fully explain Bake stage in pipeline stages ref.

avh121:


reference/pipeline/stages/index.md
```diff
@@ -13,10 +13,24 @@ sidebar:
...
 
+Note that when you're creating a pipeline, you probably won't see every stage
+that's listed here: you'll only see the stages that Spinnaker supports on your
+provider.
+
 ## General
 
 ### Bake
-Bake an image in the specified region.
+Bake an image from the specified packages. _Baking_ is the process of
```
lwander: We also use "bake" to describe creating a manifest with the "Bake (manifest)" stage

reference/pipeline/stages/index.md
```diff
@@ -13,10 +13,24 @@ sidebar:
...
+creating a machine image. Spinnaker provides default
```
lwander: I might call out that Spinnaker happens to use Packer to bake images, rather than assume that it's known why default Packer templates are provided.

reference/pipeline/stages/index.md
```diff
@@ -13,10 +13,24 @@ sidebar:
...
+[Packer templates](https://www.packer.io/intro/) and base machine images in
+order to get you started, but see the
+[bakery configuration guide](/setup/bakery/) if you want to customize your bake
+process.
+
+Note that Spinnaker looks for a previously-created image before baking a new
+one. By default, Spinnaker skips the  **Bake** stage if an equivalent image
```
lwander: These semantics are a little tricky, and based on how names are generated for the image that will be baked. I think either @duftler or @ezimanyi can shed some insight

ezimanyi: Effectively we generate a unique key for each bake based on the parameters (versioned package, base os, etc.).  In principle this means that if either the packages have changed (because of a new build, etc.) or if the paramters of the bake stage have changed (ie, using a new base image) we should trigger a new bake.  The details are here: https://github.com/spinnaker/rosco/blob/12add382dcc58f9bbbf4cce37ff49d4744c58fd8/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy#L78

reference/pipeline/stages/index.md
avh121: Hm, understood. What if we changed this to just say "Create a machine image from the specified packages" without defining baking specifically, since it has two meanings? (Or perhaps @dorbin has a better wordsmithing suggestion?) 

I didn't want to talk about the Bake (Manifest) stage here because it felt too much like a tangent, but am open to doing that if you think clarification about the differences would be helpful.

reference/pipeline/stages/index.md
avh121: Thanks for the clarification :) I've updated the text to talk about the process that you just described.

reference/pipeline/stages/index.md
lwander: Totally agree about it being a tangent -- I was just worried that our definition here prevents us from defining "Bake (Manifest)" later

avh121: Another option: we could also just say 'Baking here refers to the process of creating a machine image', and then in "Bake (Manifest)" later we can say 'Baking here refers to the process of creating a manifest'. 

What do you think makes the most sense?

ezimanyi (approved):
:+1:

reference/pipeline/stages/index.md
```diff
@@ -13,10 +13,27 @@ sidebar:
...
+[Hashicorp's Packer](https://www.packer.io/intro/). Spinnaker provides default
+[Packer templates](https://www.packer.io/docs/templates/index.html) and base
+machine images in order to get you started, but see the
+[bakery configuration guide](/setup/bakery/) if you want to customize your bake
+process.
+
+Note that Spinnaker skips the bake process if it detects that a new bake is
+unnecessary. Spinnaker generates a unique key for each bake based on the bake
+stage parameters: base OS, versioned packages, etc. If either the packages or
+the bake stage parameters have changed, Spinnaker should trigger a new bake.
```
dorbin: s/should trigger/triggers/

Unless, @lwander , is there some doubt that the new bake'll get triggered upon a change?

reference/pipeline/stages/index.md
dorbin: :+1: to Nora's rewrite&mdash;define bake implicitly by summarizing the stage behavior. There's still a chance of ambiguity between Bake and Bake (Manifest), but I don't think it's _that_ ambigious.

reference/pipeline/stages/index.md
```diff
@@ -13,10 +13,27 @@ sidebar:
 This article describes the currently-supported stages that you can add to your
 Spinnaker pipelines.
 
+Note that when you're creating a pipeline, you probably won't see every stage
+that's listed here: you'll only see the stages that Spinnaker supports on your
+provider.
```
dorbin, line 18: Are there other factors that determine which stages populate the list, besides provider? 
@lwander

reference/pipeline/stages/index.md
lwander: :+1: as well

reference/pipeline/stages/index.md
lwander: I think "triggers" is best

reference/pipeline/stages/index.md
lwander: None that are relevant here - I think some are hidden by "netflixMode", a feature flag, but that's arguably not important.

dorbin (approved):
Approved, after that one comment.

PR #854, opened 2018-06-14
fix(pipeline stages): expand canary analysis description.

avh121:


reference/pipeline/stages/index.md
```diff
@@ -37,7 +37,10 @@ the default behavior and re-bake your image each time the pipeline runs, select
 
 ### Canary Analysis
 Run [automated canary analysis](/guides/user/canary/) against the deployment
-before fully deploying.
+before fully deploying. Note that this stage _only_ handles analysis; you're
+responsible for provisioning and cleaning up your canary instances elsewhere.
```
dorbin: Does elsewhere mean in another (downstream) stage?

reference/pipeline/stages/index.md
```diff
@@ -37,7 +37,10 @@ the default behavior and re-bake your image each time the pipeline runs, select
...
+For a step by step explanation of how to set up a canary analysis stage see the
```
dorbin: step-by-step

dorbin (commented):
Couple of inline comments. 
Plus: it might be good to get the name Kayenta in there somehow. This is partly because all the press for this feature has treated Kayenta as a product, practically, rather than just a microservice that supports ACA.

reference/pipeline/stages/index.md
duftler: Could drop elsewhere and add another sentence with something like: "This is typically done within the same pipeline via stages surrounding the Canary Analysis stage."

avh121: Yeah, Matt's wording is much better -- updated it to that.

dorbin (approved):
:+1:

PR #858, opened 2018-06-16
feat(webhook-stages): add docs on preconf webhooks

ethanfrogers:
adds documentation on preconfigured webhook stages with examples

@lwander @dorbin PTAL


lwander (approved):
LGTM technically, will defer to @dorbin

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+    method: GET
+    url: https://api.example.com
+```
+
+Preconfigured webhook stages support a variety of options, most of which are available via the UI when using the vanilla Webhook stage. Once a property is set within a preconfigured webhook stage, users will not be allowed to override that settings via the UI. The following basic properties can be set within your configuration. More advanced properties will be covered in other sections.
+
+* `enabled` - whether the stage will appear in the UI
+* `label` - the human readable name of the stage
+* `description` - a human readable description of what the stage is used for
+* `type` - a _unique_ key used to identifty the stage type within a pipeline
```
dorbin: Unique among all custom webhook stages?  If so, I'd recommend you change the `type` in the example above from `customWebhook` to seem like  a more specific example.

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+* `url` - the url for the webhook
+* `customHeaders` - any headers needed for your webhook. ex. API tokens.
```
dorbin: These are headers on the http request?

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+* `method` - HTTP method used for the webhook.
+* `payload` - the JSON payload 
+
```
dorbin: I see all of these represented in the above example, except for two. So clearly `customHeaders` and `payload` are optional. Are any other fields optional. And if so, are there defaults. For example do we assume a default for `method` or is it required?

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+The value of these properties can be accessed via SpEL within various attributes of the webhook configuration. For example, if we built a stage for updating Github statuses of a commit, we could use the above property and reference it's value by using SPeL within our webhook URL.
+
+```yaml
+...
+url: https://api.github.com/repos/spinnaker/my-repo/statuses/${parameterValues['gitCommit']}
+...
+```
+
+Webhook properties are rendered in the UI as input fields.
```
dorbin: Would it be possible to include a screen shot showing this same (`gitCommit`) property rendered in the UI as an input field?

_data/navigation.yml
```diff
@@ -277,6 +277,11 @@ guides:
         url: /guides/developer/kork-library/
       - title: Kubernetes CRD extensions
         url: /guides/developer/crd-extensions/
+  - title: Admin How-Tos
```
dorbin: Kinda struggling with "Admin" versus "Operator" here. Operator seems to be used more in the docs to refer to this role, though you're putting this in a directory that's called /admin/ and which was clearly intended for this sort of information. 

Thoughts?

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
+---
+layout: single
+title:  "Preconfigured Webhook Stages"
```
dorbin: How 'bout "Customer" instead of "Preconfigured"?  It sounds like that's what this feature is--custom webhook stages; am I understanding correctly?

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+sidebar:
+  nav: guides
+---
+
+Preconfigured webhooks provide a simple, yet powerful, way of adding custom stages to Spinnaker. These stages are typically used to make quick API calls to an external system as part of a pipeline. Instead of extending the various components through code, users can simply add configuration to Orca for these stages. They appear in the UI as if they were a native stage.
+
+
+## Creating a Preconfigured Webhook Stage
```
dorbin: "Configure a custom webhook stage", maybe?

guides/admin/preconfigured-webhooks/index.md
```diff
@@ -0,0 +1,143 @@
...
+* `enabled` - whether the stage will appear in the UI
+* `label` - the human readable name of the stage
+* `description` - a human readable description of what the stage is used for
+* `type` - a _unique_ key used to identifty the stage type within a pipeline
+* `url` - the url for the webhook
+* `customHeaders` - any headers needed for your webhook. ex. API tokens.
+* `method` - HTTP method used for the webhook.
+* `payload` - the JSON payload 
+
+## Preconfigured Webhook Properties
```
dorbin: The webhooks are preconfigured, per this feature, right? Not the properties? The wording of this heading sounds like the properties are preconfigured. 

Possible wording (and again, this assumes that I'm understanding the feature right): 
"Configure properties for custom webhooks"

dorbin (changes requested):
Thanks for the doc, Ethan!

I have a few comments in line. 

One general editing/style note: we're moving toward headings that capitalize only the first word (plus any proper nouns). The title case you're using for the article title is fine, but headings inside the article would use this "sentence case" capitalization.

ethanfrogers:
@dorbin thanks for the feedback. i'll try to get this addressed soon!

guides/admin/preconfigured-webhooks/index.md
ethanfrogers: @dorbin if the field is not set in the config it will be rendered in the UI so the user of the stage can set the values.

guides/admin/preconfigured-webhooks/index.md
ethanfrogers: There is one under `Using a preconfigured webhook`.

_data/navigation.yml
ethanfrogers: I don't have any strong feelings either way. We can change the directory/title to `Operator` is that's more consistent with the rest of the docs.

guides/admin/preconfigured-webhooks/index.md
ethanfrogers: My only hesitation is that the feature, in code, is called "preconfigured" and I want to avoid any confusion. I agree that it rolls off the tongue a bit better, though. I'll change it in this PR. We can always change it back if it confuses anyone.

ethanfrogers:
@dorbin feedback addressed!

_data/navigation.yml
dorbin: A cursory search, of spinnaker.io, for both terms seems to show that operator (etc.) is more common that administrator (etc.). So Operator it is. And I think "Operator How-Tos" is ok, to make it consistent with the other sections.

guides/admin/preconfigured-webhooks/index.md
dorbin: :+1:

_data/navigation.yml
ethanfrogers: 👍 change incoming.

guides/admin/preconfigured-webhooks/index.md
dorbin: Oops, I meant "Custom" not "Customer," but I think you got that. 

What's getting me here is that the operator is coming to this guide and finding this thing that is call "preconfigured," but he or she is the one configuring it, right? For the end user it'll be preconfigured, but so will everything else. 

I agree with your plan to change it here and worry later about confusion with what's in the code.

dorbin (approved):
Thanks, Ethan!

PR #876, opened 2018-06-25
fix(zippies): enable linking into zippy contents

### LINKED ISSUE #875: Can't deep-link to content inside zippies
One unfortunate side-effect of putting a section of content into a closed zippy is that pasting a link  to that section in the URL bar wont jump you to that spot on the page.  Here's an example link:

https://www.spinnaker.io/setup/install/environment/#required-halyard-invocations-1

Clicking this should ideally open up the "Local Git" zippy and scroll the page such that the "Required Halyard Invocations" section is on-screen.

ghost:
Leaves a zippy open if it contains an element that is being linked to by the URL's fragment identifier.  After all zippies on a page are processed this script scrolls the user to the linked element because its position may have changed due to zippies collapsing further up the page.  Finally, since the script ballooned to well over 100 lines I made it all a bit more readable.

fixes https://github.com/spinnaker/spinnaker.github.io/issues/875

### DIFF
--- a/_includes/collapsible_content_script.html
+++ b/_includes/collapsible_content_script.html
@@ -1,33 +1,129 @@
 <script>
-  $(document).ready(function() {
-    $('.begin-collapsible-section').each(function(i, el) {
-      var $start = $(el);
-      var $parent = $start.parent();
-      var $heading = $parent.next();
-      var $node = $heading.next();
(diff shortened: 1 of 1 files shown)


lwander (approved):
Nice!

PR #883, opened 2018-06-29
feat(provider/kubernetes): redirect from v1 docs

lwander:
I've made this mistake a few times when trying to guess a link


dorbin (approved):
:+1:

Issue #894, opened 2018-07-09
K8s v2 provider--GKE edition


dorbin:

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 51 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

dorbin:
Closing this issue, per #896 

Other linked issue, #940 , is still relevant.

PR #884, opened 2018-06-29
feat(provider/kubernetes): adding instructions to run/manage Spinnaker on AWS EKS

anshrma:

PR #896, opened 2018-07-09
feat(setup/kubernetes): v2 gke install steps

lwander:
https://github.com/spinnaker/spinnaker.github.io/issues/894
https://github.com/spinnaker/spinnaker.github.io/pull/884


setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,61 @@
...
+  nav: setup
+---
+
+{% include toc %}
+
+This page describes how to set up a Kubernetes cluster on
+[GKE](https://cloud.google.com/kubernetes-engine/) that Spinnaker can
+manage and deploy workloads to. The process is very simple, but some
+considerations need to be made when configuring how Spinnaker will authenticate
+against your cluster.
```
dorbin: This page describes how to set up a Kubernetes cluster on GKE to be used as a Spinnaker provider. The process is very simple, but you need to do some specific things to allow Spinnaker to authenticate against your cluster.

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,61 @@
...
+
+> Note: To manage and create clusters in a given project, you will need the
```
dorbin: s/you will need/you need/

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,61 @@
...
+> `roles.container/admin` role as described
+> [here](https://cloud.google.com/kubernetes-engine/docs/how-to/iam#predefined).
+
+# Create a cluster
+
+You can create a Kubernetes cluster on GKE using either
```
dorbin: If you don't already have a cluster for this purpose, you can create one on GKE using either...

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,61 @@
...
+Follow the instructions shown in [the official
+documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#generate_kubeconfig_entry)
+to download credentials.
+
+__Warning!__ Odds are, the credentials you've downloaded rely on calling
+`gcloud` to generate a token and authenticate against your cluster. This means
+that the user (e.g. `a.employee@example.org`) you've configured `gcloud` to
+authenticate as is making requests against the cluster, instead of a fixed
+[Kubernetes service
+account](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/).
```
dorbin: "Instead of" is a little ambiguous here.

Do you mean...
against the cluster instead of against a fixed service account
or do you mean...
the user I authenticate as instead of authenticating as a fixed service account?

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,61 @@
...
+authentication against your cluster will fail. But on the other hand, and for
+the same reasons, it complicates configuring Spinnaker because the machine(s)
+running Spinnaker will need their own [Google service
+account](https://cloud.google.com/compute/docs/access/service-accounts) that
+`gcloud` will check the permissions of to generate a token for authentication.
+Given that all pods on GKE share the same service account, granting Spinnaker
+when running in GKE permission also grants all pods running alongside Spinnaker
+permission as well. For this reason, we recommend [configuring a service
+account](/setup/install/providers/kubernetes-v2/#optional-create-a-kubernetes-service-account)
+for Spinnaker to authenticate as.
```
dorbin: This is a good thing because it prevents authenticating from a different machine that hasn't already authenticated with gcloud. But it also complicates configuring Spinnaker because each machine running Spinnaker needs its own service account, which gcloud will check the permissions of in order to generate an authenication token. 

Given that all pods on GKE share the same service account, granting Spinnaker
on GKE permission also grants permission to all pods running alongside Spinnaker.
For this reason, we recommend [configuring a service account](/setup/install/providers/kubernetes-v2/#optional-create-a-kubernetes-service-account) for Spinnaker to authenticate as.

dorbin (commented):
Some suggestions; we'll talk more tomorrow

setup/install/providers/kubernetes-v2/gke/index.md
lwander: The latter

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,62 @@
...
+title:  "Google Kubernetes Engine (GKE) Setup"
+sidebar:
+  nav: setup
+---
+
+{% include toc %}
+
+This page describes how to set up a Kubernetes cluster on
+[GKE](https://cloud.google.com/kubernetes-engine/) to be used as a Spinnaker
+provider. The process is very simple, but you need to do some specific things
```
dorbin: Is it worth noting that it's a v2 provider? It'll be obvious if the user gets here from the nav, but otherwise, maybe not?

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,62 @@
...
+[Terraform](https://www.terraform.io/docs/providers/google/r/container_cluster.html)
+work too, and can be used to automate provisioning your clusters.
+
+# Download credentials
+
+Follow the instructions shown in [the official
+documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#generate_kubeconfig_entry)
+to download credentials.
+
+__Warning!__ Odds are, the credentials you've downloaded rely on calling
```
dorbin: "Odds are"

Is it not absolute? Is it possible that the credentials I've downloaded don't rely on gcloud. 

Also, if this is accurate, it's *a little* cleaner to delete "Odds are" and just say "The credentials you've downloaded probably rely on `gcloud`..." 
(you can delete "calling" too, I think.

setup/install/providers/kubernetes-v2/gke/index.md
dorbin: Then: "...instead of a fixed Kubernetes service account making the requests."

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,62 @@
...
+`gcloud` to generate a token and authenticate against your cluster. This means
+that the user (e.g. `a.employee@example.org`) you've configured `gcloud` to
+authenticate as is making requests against the cluster, instead of a fixed
+[Kubernetes service
+account](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/).
+
+This is a good thing because it prevents authenticating from a different
+machine that hasn't already authenticated with gcloud. But it also complicates
```
dorbin: `gcloud` in backticks

setup/install/providers/kubernetes-v2/gke/index.md
```diff
@@ -0,0 +1,62 @@
...
+configuring Spinnaker because each machine running Spinnaker needs its own
+service account, which gcloud will check the permissions of in order to
+generate an authenication token.
```
dorbin: "...needs its own service account. `gcloud` checks the permissions of that service account in order to..."

dorbin (approved):

setup/install/providers/kubernetes-v2/gke/index.md
lwander: Yup (as discussed offline) this is accurate. Rewording

setup/install/providers/kubernetes-v2/gke/index.md
lwander: Will call it out just in case

PR #907, opened 2018-07-13
fix(docs/reference): Expand description of several pipeline stages.

avh121:
Specifically: Clone Server Group, Enable Server Group, Find Image From
Cluster, Jenkins, and Manual Judgment.

### DIFF
--- a/reference/pipeline/stages/index.md
+++ b/reference/pipeline/stages/index.md
@@ -53,8 +53,9 @@ your cluster is a particular size, or add a pipeline expression. See the
 information about creating and using pipeline expressions.
 
 ### Clone Server Group
-Deploy the image from the original Server Group to a new Server Group. You can
-choose to copy the capacity of the original Server Group as well.
+Copies all attributes of the original Server Group into a new Server Group (the
+image that was deployed to it, its capacity, etc). You can choose to give the
+new Server Group a different capacity than the original one if desired.
 
 ### Deploy
 Deploy the previously baked or found image using the specified deployment
@@ -81,9 +82,9 @@ to disable the Server Group which is newest, oldest, or previous
 (second-most-recently deployed) when this stage starts.
 
 ### Enable Server Group
-Enable a Server Group, which means that the Server Group resumes handling
-traffic. The configuration of your Load Balancer determines how traffic is
-routed between newly-enabled Server Groups and any existing Server Groups.
+Tell Spinnaker to resume sending traffic to the Server Group. The configuration
+of your Load Balancer determines how traffic is routed among newly-enabled
+Server Groups and any existing Server Groups.
 
 ### Find Artifact From Execution
 Find and bind an artifact from another pipeline execution.
@@ -104,7 +105,7 @@ available jobs are automatically populated in the respective drop-down menus.
 ### Manual Judgment
 Wait for the user to click **Continue** before continuing. You can specify
 instructions for how to decide whether to continue, or add input options
-that users can choose between.
+that users can choose from.
 
 ### Pipeline
 Run an existing pipeline: you can select any pipeline from this application and
--- a/reference/pipeline/stages/index.md
+++ b/reference/pipeline/stages/index.md
@@ -53,7 +53,8 @@ your cluster is a particular size, or add a pipeline expression. See the
 information about creating and using pipeline expressions.
 
 ### Clone Server Group
-Deploy a new Server Group that is a copy of the specified Server Group.
+Deploy the image from the original Server Group to a new Server Group. You can
+choose to copy the capacity of the original Server Group as well.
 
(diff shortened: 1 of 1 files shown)


reference/pipeline/stages/index.md
```diff
@@ -80,31 +81,35 @@ to disable the Server Group which is newest, oldest, or previous
 (second-most-recently deployed) when this stage starts.
 
 ### Enable Server Group
-Enable a Server Group, which means that the Server Group will will resume
-handling traffic.
+Enable a Server Group, which means that the Server Group resumes handling
+traffic. The configuration of your Load Balancer determines how traffic is
+routed between newly-enabled Server Groups and any existing Server Groups.
```
dorbin: * If possible, start the definition with something other than a verbatim repeat of the name ;)

* Is it always "resumes"? Could this be a new server group being enabled, in which case this "begins handling of server traffic."? "Begins or resumes..." ?

* Re the load balancer, are there *two or more* server groups, or just two. If there can be more than two, use "among" rather than between.

dorbin (commented):
Couple of minor edits.

reference/pipeline/stages/index.md
```diff
@@ -53,7 +53,8 @@ your cluster is a particular size, or add a pipeline expression. See the
 information about creating and using pipeline expressions.
 
 ### Clone Server Group
-Deploy a new Server Group that is a copy of the specified Server Group.
+Deploy the image from the original Server Group to a new Server Group. You can
```
duftler: This reads a bit like the only thing that is by default cloned is the image, and one can choose to _also_copy the capacity. It would be more accurate to say that all the attributes of the specified server group are cloned, and one can choose to override the capacity.

reference/pipeline/stages/index.md
```diff
@@ -80,31 +81,35 @@ to disable the Server Group which is newest, oldest, or previous
...
-use this stage.
+Run the specified job in Jenkins. You must [set up Jenkins](/setup/ci/jenkins/)
+in order to use this stage. Once Jenkins is configured, your Jenkins master and
+available jobs are automatically populated in the respective drop-down menus.
 
 ### Manual Judgment
-Wait for the user to click **Continue** before continuing.
+Wait for the user to click **Continue** before continuing. You can specify
+instructions for how to decide whether to continue, or add input options
+that users can choose between.
```
duftler: Maybe: s/between/from (since they end up presented as a set of choices)

reference/pipeline/stages/index.md
avh121: Yep, it always resumes handling traffic (I checked). Server Groups start handling traffic when they're created (when an image is deployed to them), so you would only need to explicitly enable a Server Group after it had already been disabled.

avh121:
@duftler would mind you taking another look at this?

reference/pipeline/stages/index.md
```diff
@@ -53,7 +53,9 @@ your cluster is a particular size, or add a pipeline expression. See the
 information about creating and using pipeline expressions.
 
 ### Clone Server Group
-Deploy a new Server Group that is a copy of the specified Server Group.
+Copies all attributes of the original Server Group into a new Server Group (the
+image that was deployed to it, its capacity, etc). You can choose to give the
+new Server Group a different capacity than the original one if desired.
```
lwander: It's a little more general, you can choose to override any of the properties of the original server group.

reference/pipeline/stages/index.md
```diff
@@ -80,31 +82,35 @@ to disable the Server Group which is newest, oldest, or previous
 (second-most-recently deployed) when this stage starts.
 
 ### Enable Server Group
-Enable a Server Group, which means that the Server Group will will resume
-handling traffic.
+Tell Spinnaker to resume sending traffic to the Server Group. The configuration
```
lwander, line 86: This also re-enables autoscaling on GCE & AWS

reference/pipeline/stages/index.md
```diff
@@ -80,31 +82,35 @@ to disable the Server Group which is newest, oldest, or previous
...
-use this stage.
+Run the specified job in Jenkins. You must [set up Jenkins](/setup/ci/jenkins/)
+in order to use this stage. Once Jenkins is configured, your Jenkins master and
+available jobs are automatically populated in the respective drop-down menus.
 
 ### Manual Judgment
-Wait for the user to click **Continue** before continuing.
+Wait for the user to click **Continue** before continuing. You can specify
+instructions for how to decide whether to continue, or add input options
+that users can choose from.
```
lwander: I would call out that these input options can be used to determine pipeline behavior in downstream stages, using a "Check preconditions" stage as an example.

lwander (approved):

PR #910, opened 2018-07-16
feat(code): clip code that overflows and add scrollbar

ghost:
Before:

screen shot 2018-07-16 at 9 07 45 am

After:

screen shot 2018-07-16 at 9 08 00 am

Scrolling:

screen shot 2018-07-16 at 9 08 12 am

I've checked that the behaviour matches in Firefox.


dorbin (approved):
Excellent! Thanks Scott.

PR #912, opened 2018-07-17
docs(pipeline expressions): major refactor of current guide.

avh121:
- Split pipeline expressions guide into reference & overview docs
- Updated the screenshots


reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
...
+layout: single
+title:  "Reference"
+sidebar:
+  nav: reference
+---
+
+{% include toc %}
+
+Pipeline expressions allow you to reference arbitrary values about the
+state of your system in the execution of your pipelines. Each pipeline
```
dorbin: How 'bout move this 2nd sentence under the Syntax elements heading, and before the Code subhead?

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
...
+  * `scmInfo.sha1` returns the hash of the last build
+  * `scmInfo.branch` returns the branch name
+* `deployedServerGroups`: refers to the server group that was created by the
+  last deploy stage. It should look something like:
+  `{"account":"my-gce-account",
+    "capacity": { "desired":1.0,"max":1.0,"min":1.0},
+    "region":"us-central1",
+    "serverGroup":"gcespin-dev-v005"}`
+
+## Helper functions
```
dorbin: This list occurs in /guides/ too. Recommend changing that so all that's in /guides/ is a sentence describing what help functions to, generally, and linking to this reference.

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
+---
+layout: single
+title:  "Reference"
```
dorbin: "*Pipeline Expressions* Reference"?

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
...
+* [UUID](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html)
+
+## Source code
+
+Source code for the expression language is in the [spinnaker/orca
+repository](https://github.com/spinnaker/orca), mostly in the
+[ContextParameterProcessor class](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java).
+
+
+## Pipeline expression language reference
```
dorbin: I feel like "pipeline expression language reference" is what we're reading already. Is there some other way you can refer to the SpEL reference?

dorbin (approved):
A few minor points

reference/pipeline/expressions/index.md
avh121: It actually doesn't :) But you do have to explicitly load the diff to see it in this review -- I made so many changes to /guides/user/pipeline/expressions that (weirdly) github isn't displaying it automatically, which makes this review a little confusing.

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
+evaluates to _Hello-randomString-world_.
 
-When your stage executes, you will see a dropdown with the provided choice. 
+## How do I use a pipeline expression?
 
-![](13.png)
+### Where can I use pipeline expressions?
 
-In subsequent stages, you can reference the choice provided via the [#judgment( stage Name ) ](#judgmentstring) helper method `${ #judgment( 'manual judgment stage name' )}`. 
+You can use pipeline expressions in any place in the UI where you can enter
```
dorbin: delete the first "in" here.

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
+free-form text, with the exception of the pipeline configuration stage. If you
```
dorbin: I would capitalize Configuration here, because the stage name is capitalized in the UI. And again below where you mention the Configuration stage.

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
+the helper functions that are available.
 
-`${ expression here }`
+![](images/helper-func-autocomplete.png)
 
-You can interpolate expressions and strings, so
+### Helper properties
 
-`${expressionA}-randomString-${expressionB}` 
+_Helper properties_ are variables which refer to global information about the
```
dorbin: Link _Helper properties_ here to the reference doc?

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
+current pipeline execution. For example, `execution` is a variable which refers
+to the current pipeline execution, whereas `trigger` lets you access information
+about the pipeline trigger. For example, if you've set up a Jenkins trigger and
+want to know which build triggered the current pipeline, you could access that
+information with the expression `${trigger["buildInfo"]["number"]}`. For a
+complete list of available properties, see the reference guide.
```
dorbin: Then you can probably delete this last sentence because you've linked to the reference already?

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
 
-will evaluate to `Kimmy-randomString-Schmidt` when expressionA is `Kimmy` and expressionB is `Schmidt`.
+Typing a helper property into an expression displays a list of the values that
```
dorbin: Maybe move this sentence to after the paragaraph below: " To quickly list available helper properties..."

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
 
-You can set a default value within expressions by using `?:`. For example, `${ parameters.region ?: 'us-east-1'}`.
-
-###Filtering maps#
-
-The expression language allows you to filter maps by using the `.?` expression. For example, to filter a list of stages by type, you can use the expression `${ execution.stages.?[ type == 'bake' ] }`
+### #root
 
-###Lists as parameters
+`#root` returns the current stage context -- the _context_ map contained in each
```
dorbin: s/ -- /&mdash;/

guides/user/pipeline/expressions/index.md
```diff
@@ -10,492 +10,356 @@ redirect_from:
...
-![](23.png)
+Note that [Gate](https://github.com/spinnaker/gate) is the microservice that
+provides access to the Spinnaker API. If you're accessing Spinnaker via the
+default `localhost:9000`, make sure you're sending your API requests to Gate
+(defaults to `localhost:8084`).
 
-Similarly, you can see that when I chose 'cleaning up', that pipeline branch is the one that runs. 
+Note also that pipeline expressions may not behave exactly the same against the
+testing endpoint and in your pipeline. In your pipeline, each expression is
+anchored to the particular stage in which it was declared. However, when using
```
dorbin: Is "anchored" the right word here? 

"Each expression runs against the particular stage..." maybe?

avh121: I'm not sure. Do you think "runs against" conveys that the scope of what is accessible in the expression is dependent on the stage? e.g., you can't access values of future stages, and there are specific ways to access the value of the current stage. I'm definitely not tied to "anchored," but communicating that is why I picked it.

dorbin: "is limited in scope to..." ?

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
...
+
+Similarly, you can call static methods using the syntax
+`T(fully.qualified.class.name).methodName()`. For example, to calculate the date
+5 days from now, you can call:
+`${T(java.time.LocalDate).now().plusDays(5).toString()}`.
+
+### Comparisons
+
+You can use relational operators to compare values in an expression, such as
+`${instance["size"] > 400}` or `${parameters["runCanary"] == true}`.
```
travisreed-wf: I believe that this curly brace needs to be moved.  `${parameters["runCanary"]} == true`
I tried this format and the expression always evaluated to false.

See
https://github.com/spinnaker/spinnaker.github.io/pull/913

travisreed-wf: It looks like this syntax issue may exist in many of these examples.

tomaslin: It is not evaluating to true because it is a string, so you would need to surround the true part in quotes

tomaslin: Or use the #toBoolean helper function

guides/user/pipeline/expressions/index.md
tomaslin: Each expression is evaluated against the context of the particular stage ...

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,207 @@
...
+`#functionName(params)`. For example, `#fromUrl("http://www.netflix.com")`. See
+the full list of functions [here](#helper-functions).
+
+### Lists
+
+You can also use lists in your expressions. Spinnaker always provides a list
+called _stages_, which you can use to access each stage by index:
+`${execution["stages"][0]}` returns the value of the first stage in your
+pipeline. Keep in mind, though, that if you change the order in which your
+stages execute in the pipeline, the value of `stages[i]` will also change.
```
robzienert: Is this implying that SpEL expressions can change an execution and have its changes persisted? May want to reword this?

avh121: It was trying to imply the reverse -- if you rearrange your pipeline it may break your existing pipeline expressions. Is the current wording clearer?

reference/pipeline/expressions/index.md
avh121: Thanks for catching that! I've updated this to be `${parameters["runCanary"] == "true"}`, and added a note about this.

reference/pipeline/expressions/index.md
robzienert: 🥇 much clearer, thanks!

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+{% include toc %}
+
+Pipeline expressions allow you to reference arbitrary values about the
+state of your system in the execution of your pipelines. For more information
+about how pipeline expressions work, see the
+[overview](/guides/user/pipeline/expressions/).
+
+## Syntax elements
+
+Each pipeline expression is made of `$` followed by opening/closing brackets:
```
tomaslin, line 17: I think it is worth noting here that expressions cannot be nested. A common mistake we see is people trying to use  ${ something ${ something else } } and not having their expressions evaluated.

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+the full list of functions [here](#helper-functions).
+
+### Lists
+
+You can also use lists in your expressions. Spinnaker always provides a list
+called _stages_, which you can use to access each stage by index:
+`${execution["stages"][0]}` returns the value of the first stage in your
+pipeline.  Keep in mind, though, that if you change the order in which your
+stages execute in the pipeline (separately from your expression), the value of
+`stages[i]` will also change.
```
tomaslin: Worth noting that this is for illustration purposes only and the less fragile approach is to use the #stages( 'stagename' ) helper

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+2. Click on _Details_ to expand the pipeline.
+3. Click on the source link under the pipeline execution, which opens a new tab
+    containing the JSON details of your pipeline execution.
+    _[[screenshot](https://screenshot.googleplex.com/U1rVRjdiqKC.png)]_
+
+Note that while Spinnaker supports both dot (`map.value`) and square bracket
+(`map["value"]`) notation, we recommend using square brackets. That is, prefer
+`trigger["properties"]["value"]` instead of `trigger.properties.value`. There
+are a few places where using dot notation produces unexpected results, such as
+after a filter operation or when getting nested JSON values from an URL.
```
tomaslin: Another reason to recommend the square bracket notation is to deal with spaces and special characters such as dots.

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+You can filter maps using `.?`. For example, to filter a list of stages by type,
+you would use `${execution["stages"].?[type == "bake"]}`.
```
tomaslin: The result of such filtering is a list, so to access an element it should be referenced as such...

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+[Helper properties](/guides/user/pipeline/expressions/#helper-properties) are
+attribute shortcuts that you can use in Spinnaker. The specific properties are:
+
+* `execution`: refers to the current pipeline execution.
+* `parameters`: refers to pipeline parameters.
+* `trigger`: refers to the pipeline trigger.
+* `scmInfo`: refers to the git details of either the trigger or the most
+recently executed Jenkins stage.
+  * `scmInfo.sha1` returns the hash of the last build
```
tomaslin: change to git commit hash

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+Converts a JSON String into a Map that can then be processed further.
+
+### #fromUrl(String)
+
+Returns the contents of the specified URL as a String.
+
+### #jsonFromUrl(String)
+
+Retrieves the contents of the given URL and converts it into either a map or a
+list. You can use this to fetch information from the Spinnaker API or other
```
tomaslin: This is technically wrong as most of the time the Spinnaker API would be authenticated and we don't supply additional mechanisms to add authentication. Probably change to `unauthenticated URL endpoints`

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+sources.
+
+### #judgment(String)
+
+Returns the selected judgment value from the Manual Judgment stage whose name
+matches the input string. Note that `#judgment` is case sensitive:
+`${#judgment("my manual judgment stage")}` returns an error if your stage is
+named _"My Manual Judgment Stage"_.
```
tomaslin: This function is also aliased to the alternative spelling #judgement

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+### #propertiesFromUrl(String)
+
+Retrieves the contents of the properties file at the given URL and converts it
```
tomaslin: Would link to java properties format here or call out that this is the expected format

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+into a map. You can use this to fetch information from Jenkins properties files
+or other similar endpoints.
+
+### #stage(String)
+
+A shortcut to get the stage by name. For example, `${#stage("Bake")}` allows
+you to access your Bake stage. Note that `#stage` is case sensitive: if your
+stage is actually named "_bake_", `${#stage("Bake")}` will not find it. Remember
+that the values for the stage are still under the context, so you would access a
```
tomaslin: under the map called context?

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+Converts a value to a floating point number.
+
+### #toInt(String)
+
+Converts a value to an integer.
+
+### #toJson(Object)
+
+Converts a map into a JSON string.
```
tomaslin: This converts an arbitrary object so not necessarily a map. It can be the results of a jsonFromUrl for example. It is useful to pass more complex data structures into downstream systems, such as a Jenkins or Run Job stage parameter

reference/pipeline/expressions/index.md
avh121: Could you expand on what you mean by this? (e.g., what does it mean to deal with spaces?)

reference/pipeline/expressions/index.md
avh121: An arbitrary JSON object, right? What are the limits of what it can parse? (i.e., I assume I couldn't pass in an arbitrary Java object?)

reference/pipeline/expressions/index.md
avh121: Do you think that's necessary to call out here? Seems like something that people will discover on their own if that's how they spell it.

tomaslin: It's not clearly discoverable

avh121:
@jeyrschabu mind taking a look at this?

reference/pipeline/expressions/index.md
jeyrschabu: Worth mentioning that `deployedServerGroups` can also be used as a function: `${#deployedServerGroups("Deploy in us-east-1")}`. Takes the name of the deploy stage.

reference/pipeline/expressions/index.md
jeyrschabu: Would recommend using brackets for non-alphanumeric  properties

jeyrschabu (approved):
Looks good. Didn't go through all the screenshots, do we have a screenshot highlighting how to access properties from an execution json?

avh121:
@jeyrschabu yep, the screenshot titled "execution-json.png" shows how to get to the execution JSON, and the process is described in the "how do I test a pipeline expression?" section of the guide (guides/user/pipeline/expressions/index.md, as opposed to the reference file).

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,230 @@
...
+`${ }`. Within these brackets, you can add arbitrary expressions to access and
+modify the details of your pipeline during pipeline execution.
+
+Note that expressions can't be nested: `${ expression1 ${expression2} }` won't
+be evaluated.
+
+### Code
+
+Spinnaker allows you to execute Java code within a pipeline expression. This can
+be useful for string manipulation or to build more complex expressions than
```
tomaslin: `build more complex expressions` feels like the wrong term here. I suggest replacing with `useful string manipulation or more advanced custom logic`

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,230 @@
...
+would otherwise be possible. For security reasons, you can only use whitelisted
+Java classes. You can find the full list of whitelisted classes
```
tomaslin: call methods in a list of whitelisted Java classes

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,230 @@
...
+[here](#whitelisted-java-classes).
+
+You can use methods available to the whitelisted classes without importing them.
```
tomaslin: this line doesn't really make sense. There isn't a syntax for importing a class in the expression language. I would eliminate it because it is confusing

tomaslin: `You can use methods directly on existing map values based on their Java class`, is probably a better way of saying this

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,230 @@
...
+
+You can filter maps using `.?`. For example, to filter a list of stages by type,
+you would use `${execution["stages"].?[type == "bake"]}`. The result is a
+list, and can be accessed as such: `${execution["stages"].?[type == "bake"][0]}`
+returns the first bake stage in your pipeline.
+
+### Math
+
+You can use arithmetic operations in your expressions, such as
+`${trigger["buildInfo"]["number"] * 2}`.
```
tomaslin: would add a line here about helper methods such as #toint needed to convert type

reference/pipeline/expressions/index.md
```diff
@@ -0,0 +1,230 @@
...
+Strings evaluate to themselves: `${"This is a String."}` becomes "This is a
+String."
+
+## Helper properties
+
+[Helper properties](/guides/user/pipeline/expressions/#helper-properties) are
+attribute shortcuts that you can use in Spinnaker. The specific properties are:
+
+* `execution`: refers to the current pipeline execution.
+* `parameters`: refers to pipeline parameters.
```
tomaslin: This is an alias to trigger.parameters. it might make sense to reference trigger.parameters more explictly in the docs

PR #913, opened 2018-07-17
Update expression image to reflect correct syntax

travisreed-wf:
The current image includes syntax that is not properly evaluated (it will always evaluate to false).

Before this change the step is always skipped:
screen shot 2018-07-17 at 6 22 50 pm

After this change, the step is only skipped when passed the appropriate parameter:
screen shot 2018-07-17 at 6 23 53 pm


emjburns:
@travisreed-wf thanks for your PR!

PR #914, opened 2018-07-17
Fixing incorrect bookmark references in K8s V2 Helm documentation

bboerst:
The documentation incorrectly references bookmarks in Helm's documentation.

https://docs.helm.sh/helm/#helm_template

...should be:

https://docs.helm.sh/helm/#helm-template (notice the 'dash' instead of the 'underscore')

This PR corrects the bookmark so that the link is no longer broken.


dorbin:
Huh. Seems to work on my machine though, in both cases. Does the link, as is, actually 404 for you? 

In any case, good catch, @bboerst . Thanks.

dorbin (approved):

PR #916, opened 2018-07-19
Enable Kayenta in quick-install.yml

### LINKED ISSUE #908: Enable Kayenta in the quick-install.yml for Kubernetes
I think that Kayenta should be enabled by default in the quick-install.yml file for Kubernetes.
We can use the provided Minio deployment as storage backend for CONFIGURATION_STORE and OBJECT_STORE.

However, we can't enable in advance the backend for METRICS_STORE as we don't know what the user will have. I don't think it's a problem, but I'd like someone else's opinion on that one.

If you are fine with that, I can push a PR.

ping @duftler

MrTrustor:
This enables Kayenta in the quick-install.yml file for Kubernetes.
It uses the provided Minio deployment as configuration and object storage.

The user has to configure a metric storage themselves.

Fixes #908

### DIFF
--- a/downloads/kubernetes/quick-install.yml
+++ b/downloads/kubernetes/quick-install.yml
@@ -286,7 +286,38 @@ data:
         google:
           enabled: false
           subscriptions: []
-
+      canary:
+        enabled: true
+        serviceIntegrations:
+        - name: google
+          enabled: false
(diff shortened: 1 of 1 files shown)


duftler (approved):
LGTM.

PR #917, opened 2018-07-19
Persist Halyard conf in quick-install.yml

MrTrustor:
With the quick-install.yml file for Kubernetes, the Halyard
configuration could be modified but was not persisted across pod
recreation.

This addresses that by mounting a new PV
for config files and initialising them from
a ConfigMap only if they don't already exist.


MrTrustor:
Hi! Gentle ping on that one.

lwander (approved):
I like this - nice work!

Issue #922, opened 2018-07-23
Jenkins CI setup guide contains unclear references to Igor and Orca


tillig:
The section in the setup guide [pertaining to Jenkins integration](https://www.spinnaker.io/setup/ci/jenkins/) ([source](https://github.com/spinnaker/spinnaker.github.io/blob/master/setup/ci/jenkins.md)) there is a section "Configuring Spinnaker" that references Igor and Orca. For example, "Enable Igor" is all the direction you get with reference to Igor.

If you're following along the setup guide, Igor isn't enabled by default and there is no other reference anywhere explaining how to do this. Looking in the default `.hal/` folder, there's no `igor.yml` anywhere in there, so the [Igor README](https://github.com/spinnaker/igor/blob/master/README.md) also doesn't provide any clarification.

The Orca reference tells me to restart Orca. What does that mean in a Kubernetes deployment? Just kill the pod?

I found my Igor answer [by digging through forums](https://community.spinnaker.io/t/spinnaker-and-jenkins-debug/314), but based on that thread it appears I'm not the only one confused. That, and I'm not entirely sure I did it right. Does it really require manual YAML hacking or is there a Halyard command I'm missing?

It would be good to add some clarification on how to enable Igor and restart Orca such that folks following along the setup guide aren't left hanging.

dorbin:
@tillig 
Did you use Halyard to configure CI? If so, could you provide the specific steps you took?

When you use Halyard to enable CI, for example...

` hal config ci jenkins enable`

...Halyard is supposed to notice that you enabled a CI service and enable Igor for you upon `hal deploy apply`. If that's not working, we'll need to have a look.

I agree though that the linked doc is unhelpful. I'll update it after this issue is resolved. 

Thanks!

tillig:
I was following [the guide I linked](https://www.spinnaker.io/setup/ci/jenkins/) and got all the way down to "Configuring Spinnaker."

Close to the first step, just after "Prerequisites," I did `hal config ci jenkins enable`. While the guide doesn't actually say when to run `hal deploy apply` I've taken to doing that reasonably often to see what changes occur.

I got down to the "Configuring Spinnaker" stage where it said "Enable Igor." I looked in my Kubernetes cluster and did not see any deployment for `spin-igor` out there after the most recent `hal deploy apply`. Many other services, no Igor. There was also no `igor.yml` in my `~/.hal` folder.

[...]

PR #923, opened 2018-07-24
docs(setup): Clarifies the need to invoke hal deploy apply in order t…

dorbin:
…o enable Igor

### DIFF
--- a/setup/ci/jenkins.md
+++ b/setup/ci/jenkins.md
@@ -6,6 +6,8 @@ sidebar:
 redirect_from: /docs/jenkins-script-execution-stage
 ---
 
+{% include toc %}
+
 Setting up [Jenkins](https://jenkins.io/){:target="\_blank"} as a Continuous
 Integration (CI) system within Spinnaker enables using Jenkins as a Pipeline
 Trigger, as well as the Run Script stage, which depends on Jenkins as a job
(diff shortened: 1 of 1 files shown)


dorbin:
@jtk54 I included you cause I made a small change to your CSRF section.

lwander (approved):

dorbin:
#922 
@tillig

setup/ci/jenkins.md
```diff
@@ -121,29 +129,37 @@ In the UI, go to `"Manage Jenkins"` >> `"Configure System"` and set your git `us
...
-```
-jenkins.enabled = true
-jenkins.masters[0].name = <jenkins_name>
-jenkins.masters[0].address = http://<jenkins_host>/jenkins Note that "/jenkins" might not be the base path, it depends on how Jenkins is configured.
-jenkins.masters[0].username = <username>
-jenkins.masters[0].password = <user_api_token>
-```
+   Because you [enabled Jenkins](#add-your-jenkins-master), Igor should be
+   enabled automatically by Spinnaker. You can confirm this by checking your
+   `~/.hal/default/service-settings/`
```
tillig, line 138: What should I see in here? Should it have created an `igor.yml`? (I still don't have one, though it appears Igor is enabled.)

dorbin: I think you would only create an igor.yml if you needed to explicitly override an igor enabled:false.

Sorry, I merged *just* before I saw this question.

setup/ci/jenkins.md
```diff
@@ -121,29 +129,37 @@ In the UI, go to `"Manage Jenkins"` >> `"Configure System"` and set your git `us
...
+   `~/.hal/default/service-settings/igor.yml`:
 
-```
-script:
-  master: <jenkins_name> # name of Jenkins master in Spinnaker
-  job: <JOB_NAME> # from Jenkins job configuration
-```
+   `enabled:true`
+
+1. In spinnaker-local.yml, set:
```
tillig, line 145: Is this needed _in addition to_ the `hal config ci jenkins master add` call that was made earlier? (Same question for the `orca-local.yml` below.)

dorbin: Yeah, that's a good point. That might be vestigial, from before halyard. I'll have a new PR ready in the morning. 
Thanks!

dorbin: @lwander

PR #924, opened 2018-07-25
fix(docs): Remove duplicate edges in dependency graph

ezimanyi:
The Orca -> Kayenta and Gate -> Kayenta edges were duplicated


ezimanyi:
Before:
image
After:
image

ezimanyi:
Also sorted the port mapping table while I was there, so it's easier to find the microservice you're looking for.

dorbin (approved):
Good catch, Eric!

PR #928, opened 2018-07-26
docs(authz/ldap): Add halyard support docs for the ldap role provider.

stewchen:
pending halyard release

### DIFF
--- a/setup/security/authorization/ldap/index.md
+++ b/setup/security/authorization/ldap/index.md
@@ -12,23 +12,8 @@ groups.
 
 ## User DNs
 
-Sample configuration:
-```yaml
-auth:
-  groupMembership:
-    service: ldap
-    ldap:
-      url: ldaps://ldap.mydomain.net:636/dc=mydomain,dc=net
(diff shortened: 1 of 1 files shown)


setup/security/authorization/ldap/index.md
```diff
@@ -12,23 +12,8 @@ groups.
...
-      userDnPattern: uid={0},ou=users
-      groupSearchBase: ou=groups
-      groupSearchFilter: (uniqueMember={0})
-      groupRoleAttributes: cn
-```
-
-When searching for a user's groups, the `userDnPattern` is used to construct the user's full
-distinguished name (DN). In the case above, the user `joe` would have a full DN of
+When searching for a user's groups, a `userDnPattern` is used to construct the user's full
+distinguished name (DN). In the case below, the user `joe` would have a full DN of
```
ttomsu, line 16: You may want to add some language about how the manager dn & password are used to connect to the server (outside of the user's login).

ttomsu (approved):

dorbin (approved):

PR #942, opened 2018-07-30
Use message format rather than manual template

tillig:
The Jinja templates [are embedded in the pub/sub mechanism](https://github.com/spinnaker/echo/blob/90f3469f85718b86513da48a6b3dc164bca359bf/echo-pubsub-google/src/main/java/com/netflix/spinnaker/echo/config/GooglePubsubProperties.java#L83) for both GCS and GCR formats. Showing a manual template here is confusing when compared to the use of `--message-format` in the [Google pub/sub trigger setup docs](https://www.spinnaker.io/setup/triggers/google/).


lwander (approved):
Nice, thanks!

PR #944, opened 2018-08-01
docs(nav): Adds pipeline stages ref. link under guides/pipeline

dorbin:


avh121 (approved):

_data/navigation.yml
```diff
@@ -240,6 +240,8 @@ guides:
                 url: /guides/user/pipeline/triggers/gcs/
               - title: Receiving Artifacts From GitHub
                 url: /guides/user/pipeline/triggers/github/
+              - title: Pipeline Stages
+                url: /reference/pipeline/stages/
```
avh121, line 244: Noticed this is under "Triggering Pipelines" but I think it should actually be its own section.

Also, I worry that it will be confusing to people to be pulled out of the "guides" nav and into the reference pages when they click on Pipeline Stages. If they navigate to Pipeline Stages and then, say, want to get back to look at the Pipeline Triggers documentation, I think it'll be very confusing that Pipeline Triggers is no longer part of the left nav. What do you think/do you think there are ways we can mitigate that?

avh121 (changes requested):

_data/navigation.yml
dorbin: Agree on point 1. Oops, sorry 'bout that.

On point 2: valid. let's re-think this. I'll close this PR and we'll do it in another after we figure it out.

PR #950, opened 2018-08-03
_json_key is never set to USERNAME

fikriauliya:
`USERNAME` variable can't be overridden (in my Mac at least. It is always set to my Mac username).


setup/install/providers/docker-registry.md
```diff
@@ -158,7 +158,7 @@ API](https://console.developers.google.com/apis/api/cloudresourcemanager.googlea
    ```bash
    hal config provider docker-registry account add my-docker-registry \
     --address $ADDRESS \
-    --username $USERNAME \
+    --username _json_key \
```
lwander, line 159: Could you remove the assignment of `USERNAME=_json_key` on line [142 above as well](https://github.com/spinnaker/spinnaker.github.io/pull/950/files#diff-3b34abaaf580b7c9636d788f0756680fR142)?

lwander (approved):
Thanks!

setup/install/providers/docker-registry.md
fikriauliya: Done

lwander (approved):
Thanks!

PR #954, opened 2018-08-07
Added command to enable http artifact support

mvillumsen:
When setting up an http artifact account I discovered that the documentation was lacking a command to enable http artifacts.


lwander (approved):
Thanks!

PR #955, opened 2018-08-07
Include "apps" apiGroup in example RBAC role

brantburnett:
Upgrade to Spinnaker 1.8.5 is causing replicasets requests to Kubernetes to use the "apps" apiGroup instead of "app".  This change grants access to this, while maintaining access to the "app" apiGroup for users of previous versions of Spinnaker.


lwander (approved):
Good catch!

PR #961, opened 2018-08-08
fix(docs): expand descriptions of pipeline stages:

avh121:
* Find Artifact From Execution
* Find Image From Tags
* Pipeline
* Tag Image
* Wait

### DIFF
--- a/reference/pipeline/stages/index.md
+++ b/reference/pipeline/stages/index.md
@@ -89,13 +89,13 @@ Server Groups and any existing Server Groups. Enabling a Server Group also
 re-enables auto-scaling policies, if applicable.
 
 ### Find Artifact From Execution
-Find and bind an artifact from a different execution to the current one,
-including artifacts from different Spinnaker applications. To do this, specify
-the type of artifact (GCS, GitLab, Docker, etc) and its path. You can specify
-that Spinnaker should only consider executions that meet certain criteria, such
-as ones that have completed successfully or that are currently running.
-Spinnaker will always return the artifact from the most recent execution that
-matches your criteria.
+Find and bind an artifact from a different pipeline execution to the current
+one, including artifacts from different Spinnaker applications. To do this,
+specify the type of artifact (GCS, GitLab, Docker, etc) and its name. You can
+specify that Spinnaker should only consider executions that meet certain
+criteria, such as ones that have completed successfully or that are currently
+running. Spinnaker will always return the artifact from the most recent
+execution that matches your criteria.
 
 ### Find Image From Cluster
 Find an image to deploy from an existing Cluster. Make sure that you specify the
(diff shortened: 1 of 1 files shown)


reference/pipeline/stages/index.md
```diff
@@ -89,15 +89,23 @@ Server Groups and any existing Server Groups. Enabling a Server Group also
 re-enables auto-scaling policies, if applicable.
 
 ### Find Artifact From Execution
-Find and bind an artifact from another pipeline execution.
+Find and bind an artifact from a different execution to the current one,
```
dorbin: different *pipeline* execution?

reference/pipeline/stages/index.md
```diff
@@ -89,15 +89,23 @@ Server Groups and any existing Server Groups. Enabling a Server Group also
...
+including artifacts from different Spinnaker applications. To do this, specify
+the type of artifact (GCS, GitLab, Docker, etc) and its path. You can specify
+that Spinnaker should only consider executions that meet certain criteria, such
+as ones that have completed successfully or that are currently running.
+Spinnaker will always return the artifact from the most recent execution that
+matches your criteria.
```
dorbin: Can you show an example? Specifically, I'm curious about what the path looks like.

reference/pipeline/stages/index.md
```diff
@@ -89,15 +89,23 @@ Server Groups and any existing Server Groups. Enabling a Server Group also
...
 
 ### Find Image From Cluster
 Find an image to deploy from an existing Cluster. Make sure that you specify the
 Cluster and Server Group such that there is exactly one match, or this may
 behave in unexpected ways.
 
 ### Find Image From Tags
-Find an image to deploy from tags.
+Find the newest image that matches all specified tags. Tags can only contain
```
dorbin: [This](https://www.spinnaker.io/guides/user/tagging/) kind of tag? Or a tag from the repo? Or Docker registry?

dorbin: Ah. Tag as described below under Tag Image, right? If so, maybe link to that?

dorbin (changes requested):
A few minor points

reference/pipeline/stages/index.md
```diff
@@ -89,15 +89,23 @@ Server Groups and any existing Server Groups. Enabling a Server Group also
 re-enables auto-scaling policies, if applicable.
 
 ### Find Artifact From Execution
-Find and bind an artifact from another pipeline execution.
+Find and bind an artifact from a different execution to the current one,
+including artifacts from different Spinnaker applications. To do this, specify
+the type of artifact (GCS, GitLab, Docker, etc) and its path. You can specify
```
lwander: "path" isn't a concept for all artifacts, I think you mean "name".

lwander (approved):

reference/pipeline/stages/index.md
avh121: Expanded the tagging description to try to clarify briefly -- let me know what you think. 

We may also want to add some documentation somewhere about how Spinnaker interacts with image tags, since I don't think that exists anywhere.

reference/pipeline/stages/index.md
avh121: Lars pointed out that it should actually be the image name -- turns out "path" was GCS-specific :) I think that becomes clear enough that I think the example isn't necessary, particularly since the hover text for each type of image is pretty informative.

dorbin (approved):
Awesome!

PR #962, opened 2018-08-08
docs(canary): adds some new best practices

dorbin:


guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
 You can put a critical metric in a group of only one to ensure that if it fails,
 the whole group fails every time.
 
+## Compare canary against baseline, not against production
+
+You might be tempted to compare the canary deployment against your current
+production deployment. Instead always compare the canary against an equivalent
+baseline, with both deployed together.
```
demobox: "with both deployed at the same time" or "...using the same sizing" or "...similarly"

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+The baseline uses the same version and configuration that is currently running
+in production, but is otherwise identical to the canary:
+
+* Same time of deployment
+* Same size of deployment
+* Both receive the same amount of traffic
+
+Doing this, you control for version and configuration only, and you reduce
```
demobox: [minor] "In this way, ..."

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+factors that could affect the analysis, like the cache warmup time, the heap
+size, and so on.
+
+## Run the canary for a long-enough time
```
demobox: [minor] "...for long enough"

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+You need at least 50 data points for the statistical analysis to produce
```
demobox: Does a reader understand at this point what a "data point" is?

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+To compare the metrics between baseline and canary, Kayenta needs the exact same
+metrics for both. This means that metrics should have the same labels. Problems
+can arise if the metrics are labeled with their respective instance names.
+
+If you see that Kayenta is not actually comparing the metrics, confirm that the
+queries that it does to your monitoring system return metrics with the same
+labels.
+
+If you're using Stackdriver, you can use the Google APIs Explorer to debug such
```
demobox: Can we link to the API Explorer here?

demobox:
Possibly independent of this PR, but: worth adding a section about understanding that the purpose of a canary is not to detect improvements, but to detect _deviations_. I.e. if your app was misbehaving beforehand, a canary will not catch that if the new version misbehaves similarly.

/cc @duftler

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,112 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+analysis](#run-the-canary-for-a-long-enough-time) you have to wait a long time
+for the analysis to finish before you can refine it.
+
+Fortunately, a Canary Analysis stage can be configured to use a [retrospective
+analysis](/guides/user/canary/stage/#real-time-versus-retrospective-analysis)
+instead of a real-time analysis. This analysis is based on past monitoring data,
+without having to wait for the data points to be generated. With this mode, you
+can iterate more quickly on the development of the canary configuration.
+
+## Compare apples to apples
```
MrTrustor: I was not satisfied with this title when I wrote it. Hard to understand for people who don't speak a good enough english.
Other suggestions:
* Compare what can be compared
* Compare similar metrics

guides/user/canary/best-practices/index.md
MrTrustor: Link: https://developers.google.com/apis-explorer/#search/timeseries/m/monitoring/v3/monitoring.projects.timeSeries.list

guides/user/canary/best-practices/index.md
dorbin: Eh, I changed this a bit, but I don't think this page is the place to ensure they understand what they're getting into with metrics. That is, if they're going to get metrics from a provider and use them, I'm going to assume (on this page at least) that they understand what a time series is, or they'll realized that they're going to need to.

guides/user/canary/best-practices/index.md
dorbin: Yes, you're right about non-native speakers. I went with "Compare equivalent deployments"

guides/user/canary/best-practices/index.md
dorbin: :+1:

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+You might be tempted to compare the canary deployment against your current
+production deployment. Instead always compare the canary against an equivalent
+baseline, deployed at the same time.
+
+The baseline uses the same version and configuration that is currently running
+in production, but is otherwise identical to the canary:
+
+* Same time of deployment
+* Same size of deployment
+* Both receive the same amount of traffic
```
duftler: I'd say '..same amount and type of traffic'.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+  fails.
+
+* pass
+
+  The last canary run of the analysis must score higher than this threshold for
+  the whole analysis to be considered successful. Otherwise it fails.
+
+These thresholds are very important for the analysis to give an accurate result.
+You need to experiment with them, in the context of your own application, its
+traffic and its metrics.
```
duftler, line 64: Could add: Keep in mind that your automated canary analysis configuration will need to be refined over time. Don't think of it as something you set up once and never think about again.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+Use a variety of metrics that reflect different aspects of the health of your
+application. Use these three out of the four "golden signals," as defined in
+the [SRE Book](https://landing.google.com/sre/book/chapters/monitoring-distributed-systems.html):
+
+* latency
+* errors
+* saturation
+
+If you consider some other specific metrics critical, place them in their own
```
duftler, line 87: There is a 'criticality' setting that can be applied to individual metrics for this purpose.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+## Create a standard, resusable canary config
+
+Configuring a canary is difficult, and not every developer in your organization
+will be able to do so. Also, if you let all the teams in the org manage their
+own configs, you will likely end up with too many configs, too many metrics,
+nobody will know what is happening, and people will be afraid to change
+anything.
+
+For these reasons, it's a good idea to curate a config that all the teams can
```
duftler: s/a config/a set of configs

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+* latency
+* errors
+* saturation
+
+If you consider some other specific metrics critical, place them in their own
+group in the canary configuration. That allows you to fail the whole canary
+analysis if there is a problem with one of those specific metrics.
+
+## Create a standard, resusable canary config
```
duftler: Maybe: Create a set of standard, reusable canary configs.

Different classes of applications will have differing sets of metrics that are strong indicators of a safe release.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+It takes a long time to configure a canary analysis. It can take a long time to
+debug it too, partly because with [a long-running canary
+analysis](#run-the-canary-for-a-long-enough-time) you have to wait a long time
+for the analysis to finish before you can refine it.
+
+Fortunately, a Canary Analysis stage can be configured to use a [retrospective
+analysis](/guides/user/canary/stage/#real-time-versus-retrospective-analysis)
+instead of a real-time analysis. This analysis is based on past monitoring data,
+without having to wait for the data points to be generated. With this mode, you
+can iterate more quickly on the development of the canary configuration.
```
duftler, line 114: Could add: Make use of the Copy-to-clipboard icons in the canary stage execution details view to capture the needed timestamps.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,111 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+
+## Compare equivalent deployments
+
+To compare the metrics between baseline and canary, Kayenta needs the exact same
+metrics for both. This means that metrics should have the same labels. Problems
+can arise if the metrics are labeled with their respective instance names.
+
+If you see that Kayenta is not actually comparing the metrics, confirm that the
+queries that it does to your monitoring system return metrics with the same
```
duftler: s/does to your/executes against your

The easiest way to view the actual expanded queries is to navigate to the `Metrics` link under `Source` in the Reports view.

duftler (approved):
LGTM. Would also like @csanden to weigh in.

guides/user/canary/best-practices/index.md
```diff
@@ -16,6 +16,118 @@ critical metric fails, but the rest pass, the group gets a passing score overall
...
+* Same size of deployment
+* Same type and amount of traffic
+
+In this way, you control for version and configuration only, and you reduce
+factors that could affect the analysis, like the cache warmup time, the heap
+size, and so on.
+
+## Run the canary for enough time
+
+You need at least 50 pieces of time series data per metric for the statistical
```
MrTrustor, line 38: @jpassing 's comment from our doc:

> Say, for the sake of argument, that I am sampling every millisecond. Would not 50 datapoints be pretty insigificant in that case? During that time, the app may have only dispatched a single request -- or maybe even none. Maybe memory usage doubles every second -- 50 datapoints will not catch that.
I'd expect that you need at least maybe 50 datapoints AND during that time the app should have done some serious work AND you should run the analysis for long enough to give you some clue about stability.

MrTrustor: I think that Johannes has a point here. And that it actually depends on the monitoring system.
Unless I'm mistaken, Stackdriver won't give you more than one data point per minute in the answer, which gives you a 50m minimum for a canary run. But this is not the case for Prometheus or Datadog AFAIK, where you can have more data points. It's not uncommon for Prometheus users to have a data point every 15s.
@duftler how do we deal with that?

dorbin: @csanden WDYT?

duftler: @csanden is definitely the authority here. My point in the earlier thread was that it wouldn't affect the classification. But yes, you will be constrained by the precision of your tsdb. The classification logic should have a minimum of 50 data points for you to have confidence in the result, but the result will only be meaningful for a given use case if the baseline and canary are exposed to true production workload during the run. I agree with Johannes's point and maybe Chris can recommend some good wording here?

dorbin:
@csanden 
If you get a chance to look at this. I'd like to merge this, if possible. :)

duftler:
Let's merge this one before it gets too stale. Can always revise later.

PR #967, opened 2018-08-13
Replace description of Traffic guard + typo

tjnicholas:
I was a bit confused by the old description of a traffic guard. I believe it had an accidental negative inverting the meaning of the sentence. 

The replacement text is taken from the [web interface description](https://github.com/spinnaker/deck/blob/c1f63e879791473e86d0ecc0a316c3e94a9ba8de/app/scripts/modules/core/src/application/config/trafficGuard/trafficGuardConfig.component.html#L4-L6).


dorbin (approved):
Thanks, Tim. Good catch(es)!

@lwander  for sanity check

lwander (approved):
Thanks!

Issue #946, opened 2018-08-02
`deck-start.sh` script needs netcat


ttomsu:
Page: [/setup/install/environment/](https://www.spinnaker.io/setup/install/environment/)

Feedback:

The local git based installation threw a bunch of these at me during startup:

```
/home/ttomsu/dev/spinnaker/scripts/deck-start.sh: line 26: nc: command not found
```

Needed to `sudo apt-get install netcat` to make them go away - this should probably be added to the docs and/or corrected in the `desk-start.sh` script. (cc @lwander )

lwander:
Hey @brandonnelson3 I assigned you since you've implemented the local git installation

brandonnelson3:
deck-start.sh uses netcat as a sort of health check looking to block until the yarn server is up in LocalGit mode. We could, install netcat if its missing, or alternatively use curl if available, or just document netcat as a prereq.

This occurs on the GCP Drawfork debian base image, but does not manifest on any of the Ubuntu flavors. 

Thoughts?

lwander:
I would just document that it's needed in the installation instructions for this particular install path. I wouldn't try to install it since this could be running on a number of different systems, and we'll just be creating a headache for ourselves

ttomsu:
Given that the instructions on that page already have you install `curl`  if it isn't already available, I'd vote for just using `curl`.

dorbin:
For now, just adding the instruction to install netcat

PR #974, opened 2018-08-15
docs(setup): adds netcat dependency to local git env setup

dorbin:


ttomsu:
LGTM, but it's weird that we're using `netcat` at all instead of just `curl`.

ttomsu (approved):

dorbin:
Yeah, Brandon and I discussed it yesterday, and this addition is just to make the doc complete for the current situation. Changing the script to use curl is a [separate matter](https://github.com/spinnaker/spinnaker.github.io/issues/946).

PR #975, opened 2018-08-15
docs(gcb): Adds provisional instructions for trigging pipelines on Go…

ttomsu:
…ogle Cloud Build builds.

We obviously need to make this **way** easier, but here's how one could do pipeline triggers from GCB using GCB's pubsub topic.

cc @duftler @skim1420 

Rendered Preview: 
5xeepvcq3fx

### DIFF
--- a/setup/ci/gcb.md
+++ b/setup/ci/gcb.md
@@ -1,0 +1,71 @@
+---
+layout: single
+title:  "Google Cloud Build"
+sidebar:
+  nav: setup
+---
+
+{% include toc %}
+
+Setting up [Google Cloud Build](https://cloud.google.com/cloud-build/) as a Continuous Integration (CI)
+system within Spinnaker enables triggering pipelines when builds complete.
+
+## Prerequisites
+
+You need to have a [Google Cloud Platform](https://cloud.google.com) project with the
+[Cloud Build API](http://console.cloud.google.com/apis/library/cloudbuild.googleapis.com) enabled.
+You can enable the API with the following `gcloud` command:
+
+```
+gcloud services enable cloudbuild.googleapis.com
+```
+
+## Configure Spinnaker to Listen for Google Cloud Build Pub/Sub Notifications
+
+Google Cloud Build sends [Build Notifications](https://cloud.google.com/cloud-build/docs/send-build-notifications)
+when the state of your build changes. You can create a pipeline trigger to invoke a pipeline based
+on the status of your build.
+
+1. Create a Subscription object for the `cloud-builds` topic in your project:
+
+    ```
+    PROJECT_ID=
+    SUBSCRIPTION_NAME=googleCloudBuilds
+
+    gcloud pubsub subscriptions create $SUBSCRIPTION_NAME \
+      --topic projects/$PROJECT_ID/topics/cloud-builds \
+      --project $PROJECT_ID
+    ```
+
+2. Configure and deploy Spinnaker via Halyard with your newly created subscription.
+
+    ```
+    hal config pubsub google subscription add $SUBSCRIPTION_NAME \
+      --project $PROJECT_ID \
+      --subscription-name $SUBSCRIPTION_NAME \
+      --message-format GCB
+
+    hal config pubsub google enable
+
+    hal deploy apply
+    ```
+
+## Configure Your Pipeline Trigger
+
+1. In your Pipeline configuration, click the **Configuration** stage on the far left of the pipeline diagram.
+
+2. Click **Automated Triggers**.
+
+3. In the **Type** field, select `Pub/Sub`.
+
+4. In the **Pub/Sub System Type** field, select `google`.
+
+5. In the **Subscription Name** field, select your `$SUBSCRIPTION_NAME` value.
+
+6. In the **Attribute Constraints** field, enter `status` in the **Key**, and `SUCCESS` (all upper case) in the **Value** field.
+
+7. In the **Payload Constraints** field, you can enter any of the top-level fields from the
+[Build object documentation](https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#resource-build)
+as the key, and a Java regular expression as the value. For example, to match on an image produced,
+you can use the key `images` and the value `gcr\.io/my-project-id/my-application:latest` (replacing
+`my-project-id` and `my-application` with appropriate values).
--- a/setup/ci/index.md
+++ b/setup/ci/index.md
@@ -22,6 +22,7 @@ independently of one-another, enumerated below.
 
 These are the CI systems currently supported by Spinnaker:
 
+* [Google Cloud Build](/setup/ci/gcb/)
 * [Jenkins](/setup/ci/jenkins/)
 * [Travis CI](/setup/ci/travis/)
 * [Wercker](/setup/ci/wercker/)
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -142,10 +142,12 @@ setup:
         children:
           - title: Set Up Continuous Integration
             url: /setup/ci/
-          - title: Travis CI
-            url: /setup/ci/travis/
+          - title: Google Cloud Build
+            url: /setup/ci/gcb/
           - title: Jenkins
             url: /setup/ci/jenkins/
+          - title: Travis CI
+            url: /setup/ci/travis/
           - title: Wercker
             url: /setup/ci/wercker/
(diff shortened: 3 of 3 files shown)


skim1420:
Let's hold on this doc until the echo NPE bug on no artifacts is addressed.

ttomsu:
Echo fix is here: https://github.com/spinnaker/echo/pull/315

setup/ci/gcb.md
```diff
@@ -0,0 +1,66 @@
...
+
+4. In the **Pub/Sub System Type** field, select `google`.
+
+5. In the **Subscription Name** field, select your `$SUBSCRIPTION_NAME` value.
+
+6. In the **Attribute Constraints** field, enter `status` in the **Key**, and `SUCCESS` (all upper case) in the **Value** field.
+
+7. In the **Payload Constraints** field, you can enter any of the top-level fields from the
+[Build object documentation](https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#resource-build)
+as the key, and a Java regular expression as the value. For example, to match on an image produced,
```
skim1420, line 69: For example, you can trigger only on successful builds with the key `status` and the value `SUCCESS`, or to match on an image produced...

setup/ci/gcb.md
```diff
@@ -0,0 +1,66 @@
...
+
+2. Click **Automated Triggers**.
+
+3. In the **Type** field, select `Pub/Sub`.
+
+4. In the **Pub/Sub System Type** field, select `google`.
+
+5. In the **Subscription Name** field, select your `$SUBSCRIPTION_NAME` value.
+
+6. In the **Attribute Constraints** field, enter `status` in the **Key**, and `SUCCESS` (all upper case) in the **Value** field.
```
demobox, line 65: Perhaps add something like "This would trigger on **all** successful Cloud Build builds. To trigger only on builds for a particular repo, or producing a particular image, use **Payload Constraints**."?

setup/ci/gcb.md
```diff
@@ -0,0 +1,66 @@
...
+
+7. In the **Payload Constraints** field, you can enter any of the top-level fields from the
+[Build object documentation](https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#resource-build)
+as the key, and a Java regular expression as the value. For example, to match on an image produced,
+you can use the key `images` and the value `gcr\.io/my-project-id/my-application:latest` (replacing
+`my-project-id` and `my-application` with appropriate values).
```
demobox, line 71: [minor] Perhaps something like "To trigger only on builds that publish a container image with the `latest` tag to the `gcr.io/my-project/my-application` repository, add a Payload Constraint with key `images` and value `gcr\.io/my-project-id/my-application:latest` (note that the `.` needs to be escaped since the constraint value is interpreted as a regular expression)"?

/cc @dorbin

setup/ci/index.md
```diff
@@ -22,6 +22,7 @@ independently of one-another, enumerated below.
 
 These are the CI systems currently supported by Spinnaker:
```
demobox, line 23: @dorbin What are our thoughts generally on what it means when we say a "CI system is supported by Spinnaker"? Does it mean "can be integrated with in _some_ way"? Or do we consider support for a CI system to mean a) ability to trigger on completion of CI builds and b) ability to call the CI system *from* a Spinnaker pipeline via a stage?

setup/ci/gcb.md
```diff
@@ -0,0 +1,66 @@
...
+
+    ```
+    hal config pubsub google subscription add $SUBSCRIPTION_NAME --project $PROJECT_ID --subscription-name $SUBSCRIPTION_NAME
+
+    hal config pubsub google enable
+
+    hal deploy apply
+    ```
+
+## Configure Your Pipeline Trigger
```
demobox, line 53: Any way we could extract artifact information from the trigger?

/cc @ezimanyi

lwander: Yes with the right jinja template. If argo changes their payload to include "digest" (important for reproducible deployments) we will need to change this template.

skim1420: I'm working with Clod Build team to establish payload. We'll package a default jinja template with echo thereafter.

ttomsu: Here's the schema of the message payload - what fields do we need from it to make artifacts work?

https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#Build

ttomsu: @lwander and I spoke with David, and he pointed out where the digest is contained in the pubsub message, so Lars is going to proceed with adding the jinja template.

setup/ci/index.md
dorbin: A good question. And I'll evade it for the moment by saying we're not enabling CI for Build in the way we are for Jenkins, Travis, and Wercker. There isn't `hal config ci google enable`.

I think this doc should go in /guides/user/pipeline/triggers/, for now at least. *But*, we should probably include a note on this page saying you can use Build with Spinnaker.  My two ¢.

demobox: > I think this doc should go in /guides/user/pipeline/triggers/

Most CI systems provide triggers too, though - what then **is** our definition of what a "CI system" adds to Spinnaker? Should Jenkins, Travis etc. (if they can function as triggers) also be mentioned in the [trigger guide](https://www.spinnaker.io/guides/user/pipeline/triggers/)?

lwander: I would say GCB should be classified as a _supported_ CI system since we can not only trigger off of builds, but also consume artifacts produced by it.

If we want to add `hal` commands to give it the appearance of a typical CI system, that's definitely worth doing and thinking over as we simplify our artifact flows to abstract event streams into trigger types (i.e. a user should be thinking of getting a docker image from GCB, not translating a pub/sub message from a certain subscription).

ttomsu:
+1, but I don't see how we don't expose the underlying implementation here.
Lars, would you be opposed to `hal config ci gcb --subscription ____
--projectID ____ --jsonPath ___`?

lwander:
That sounds good - but let's wait to implement that until we decide how we refactor triggers. For now I think it's OK to treat it as pub/sub until we have something better. In the end, we will always leave pub/sub as an escape hatch for more advanced use cases.

demobox:
@dorbin Quick ping on this? spinnaker/echo#319 has been merged now, and it would be great if searching for "Cloud Build" in the Spinnaker site would explain how to use this new integration

setup/ci/gcb.md
```diff
@@ -0,0 +1,66 @@
...
+1. Create a Subscription object for the `cloud-builds` topic in your project:
+
+    ```
+    PROJECT_ID=
+    SUBSCRIPTION_NAME=googleCloudBuilds
+
+    gcloud pubsub subscriptions create $SUBSCRIPTION_NAME --topic projects/$PROJECT_ID/topics/cloud-builds
+    ```
+
+2. Configure and deploy Halyard with your newly created subscription.
```
dorbin: I would say "deploy Spinnaker via Halyard"

setup/ci/gcb.md
dorbin: 👍

dorbin (approved):
Couple of minor points

demobox:
> dorbin approved these changes 18 hours ago

@ttomsu @lwander Is the content still accurate?

lwander:
@demobox Yup

setup/ci/gcb.md
ttomsu: I think several of these instructions need to be updated because of artifact support from @ezimanyi . Eric, could you take a quick pass at how these instructions _should_ read now that we have first class GCB artifact support? IIRC, you mentioned something the other day about adding a specific filter, but I don't remember what it was.

lwander: We added that artifact filter when these docs were written, as far as I know we haven't done any extra work for GCB since then. But @ezimanyi can correct me

setup/ci/gcb.md
ttomsu: Done.

dorbin:
Gonna go ahead and merge this. Ultimately, we want docs in /guides/user/ for how to trigger, but in the mean time this is useful, and as @demobox points out it gets gcb into site search results.

ezimanyi:
Looks good to me!

demobox:
> Gonna go ahead and merge this

Thanks, all!

PR #976, opened 2018-08-16
Update quick-install.yml for Kubernetes to use 1.9.0

MrTrustor:
Note: this is needed for the upcoming Kayenta tutorial on cloud.google.com (see spinnaker/spinnaker#3205).


lwander (approved):
Thanks!

PR #983, opened 2018-08-22
Update Kubernetes V2 provider RBAC documentation

timReynolds:
Having followed the RBAC documentation today it's currently outdated using the default namespace and not setting enough permissions by default.


lwander (approved):
Thanks!

PR #988, opened 2018-08-23
feat(guides): Orca SQL setup

robzienert:
@asher Would appreciate you keeping me honest on the SQL settings.

Happy to move the mariadb extension repo into spinnaker org if we want, although I don't think it'll get a whole lot of activity.

Shouldn't be merged until we upstream `orca-sql`.


guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
+---
+layout: single
+title:  "Orca: Redis to SQL"
```
dorbin: Maybe swap this title with the one in `navigation.yml`. Better to have shorter titles in the nav because it looks better if they don't break a line. But the title shown on the article page can be longer and more descriptive. 

So, in nav: 
Orca: Redis to SQL

and on the page:
"Set up Orca to use SQL"

I think you can leave migration out of the title, but I don't feel strongly about it.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+Orca's execution state is stored in Redis by default, but can be configured for SQL. In this topology, Redis will still be required for the work queue.
```
dorbin: s/will still be/is still/

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+```sql
+CREATE SCHEMA `orca` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ;
+```
+
+Orca is configured to support a `service` user and a `migration` user:
+
+```sql
+GRANT SELECT, INSERT, UPDATE, DELETE, EXECUTE, SHOW VIEW ON `orca`.* TO 'orca_service'@'%'
+
+GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, REFERENCES, INDEX, ALTER, LOCK TABLES, EXECUTE, SHOW VIEW ON `orca`.* TO 'orca_migrate'@'%'
```
dorbin: Wondering where these SQL blocks go so that Orca picks 'em up.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+When Orca starts up, it will perform database migrations to ensure its running the correct schema.
+
+### Amazon Aurora
+
+At Netflix, we use Amazon Aurora MySQL and have configured it with the following settings, missing values are defaults.
```
dorbin: Putting it this way makes the whole article from Netflix POV. Not a big deal, and I think keeping the Netflix name and perspective is good. But it might work better to present it as an example, like this:
"
### Amazon Aurora example

Netflix uses Amazon Aurora MySQL and has it configured with...
"

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+- *character_set_filesystem*: `utf8mb4`
+- *character_set_results*: `utf8mb4`
+- *character_set_server*: `utf8mb4`
+- *collation_connection*: `utf8mb4_unicode_ci`
+- *collation_server*: `utf8mb4_unicode_ci`
+- *innodb_checksums*: `0`
+
+#### MariaDB
+
+We use the MariaDB JDBC driver for connecting to Aurora MySQL. The MariaDB driver is Aurora cluster aware, which takes care of automatic master failover operations. Due to licensing issues, we can't ship Orca with the MariaDB driver.
```
dorbin: And if you go that route, changing the perspective, this would be "Netflix uses..."

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Reporting on active execution metrics will be handled by SQL
+monitor:
+  activeExecutions:
+    redis: false
+```
+
+## Migrating from SQL to Redis
+
+Migrating without downtime from Redis to SQL is a three step process:
```
dorbin: s/three step/three-step/

guides/operator/orca-redis-to-sql.md
dorbin: And similarly later in the article. If it looks like it's going to be difficult or awkward to make Netflix third person rather than first person, it's no big deal if you want to keep it as is.

guides/operator/orca-redis-to-sql.md
lwander: I would think this should live in "setup" rather than "guides", but open to either. cc @dorbin

guides/operator/orca-redis-to-sql.md
lwander: I wouldn't include "at netflix, we use", rather leave it as "This has been tested against Amazon's Aurora MySQL with the following settings..."

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. Deploy Orca with the `DualExecutionRepository` writing to both Redis and SQL.
+2. Deploy a new Orca cluster with migrators enabled and queue processing disabled.
```
lwander: Is it possible to deploy only one orca that has both migrators enabled as well as queue processing to simplify the migration behavior?

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+    intervalMs: 1800000
+
+queue:
+  redis:
+    enabled: false
+
+keiko:
+  queue:
+    enabled: false
+```
```
dorbin, line 72: Is this yaml still the Netflix example, or is this a generic how-to? If still Netflix specific, can you change the heading of this section to 

### Deploying a migration cluster&mdash;a Netflix example

Again, I like using Netflix as an example throughout, but if we make Netflix third person instead of third person we can make it clearer  what parts are generic instruction and what parts are examples of how Netflix does it.

dorbin (commented):
Another great article, Rob. Thanks!

Most of my comments are about point of view, but I'm not 100% sure if I'm being clear. Feel free to reach out to me to discuss.

guides/operator/orca-redis-to-sql.md
asher: I prefer "Netflix users the MariaDB.." here.

guides/operator/orca-redis-to-sql.md
asher: Include major versions - i.e. Aurora MySQL 5.7 and perhaps note that we locally develop against or have also tested with OSS MySQL 5.7.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+- *log_throttle_queries_not_using_indexes*: `60`
+- *long_query_time*: `0.5`
+- *max_allowed_packet*: `25165824`
+- *max_binlog_size*: `134217728`
+- *query_cache_size*: `{DBInstanceClassMemory/24}`
+- *query_cache_type*: `1`
+- *read_buffer_size*: `262144`
+- *slow_query_log*: `1`
+- *sync_binlog*: `1`
+- *tx_isolation*: `READ-COMMITTED`
```
asher: I think our use of `READ-COMMITTED` is worth noting outside of this block as it is both non-default and essential. While orca attempts to set this at the session level regardless, it may be of interest to DBE's operating mysql for orca.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Aurora DB Parameters Group
+
+- *binlog_checksum*: `NONE`
```
asher: Note that if using Aurora in AWS, binlogging can be left off for single region, multi-AZ deployments, where Aurora's custom replication is preferred.

guides/operator/orca-redis-to-sql.md
dorbin: @lwander and I just had a talk about this, and his idea is to split this up into 
* Set up Orca to use SQL (in /setup/)
* Migrate existing Orca config to use SQL (here in /guides/operator/)

WDYT, @robzienert ?

guides/operator/orca-redis-to-sql.md
dorbin: This is similar to what I was saying in my comments.

guides/operator/orca-redis-to-sql.md
asher: This does work but has not been extensively tested. The migrators can be memory demanding and some users would potentially need to run with a more generous heap to successfully do all the things without impact.

guides/operator/orca-redis-to-sql.md
asher: This is a generic configuration for use with a stand-alone migration cluster.

guides/operator/orca-redis-to-sql.md
robzienert: This is something that the operator has to do to setup a database - Orca won't be doing this on behalf of the users. I'll clarify.

guides/operator/orca-redis-to-sql.md
robzienert: I can get behind that!

robzienert:
Thanks everyone for your feedback. I've finally gotten around to applying your suggestions. Please take another look, as things are quite different from the original.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+title:  "Migrate existing Orca config to use SQL"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+If you are not migrating an existing Orca deployment, refer to [Orca SQL Setup](/setup/productionize/persistence/orca-sql/) instead.
+
+## Migrating from SQL to Redis
```
avram: I think you have this to/from backwards

robzienert: Haaa. :(

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+
+If you are not migrating an existing Orca deployment, refer to [Orca SQL Setup](/setup/productionize/persistence/orca-sql/) instead.
+
+## Migrating from Redis to SQL
+
+Migrating without downtime from Redis to SQL is a three-step process:
+
+1. Deploy Orca with the `DualExecutionRepository` writing to both Redis and SQL.
+2. Deploy a new Orca cluster with migrators enabled and queue processing disabled.
+3. Once all executions have been migrated, delete migration cluster and disable `DualExecutionRepository`.
```
dorbin: You might want to link these steps to the corresponding sections below.

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+
+When `DualExecutionRepository` is running, writes will be routed to either Redis or SQL. 
+Executions will only be migrated to SQL once they've completed (either successfully or terminally): This keeps the migration story simple.
+As such, the migration agents will need to run for awhile. 
+At Netflix, we ran the migration cluster for two weeks, as we had long pipeline executions due to canaries. 
+You may only need to run the migration cluster for an hour.
+
+**NOTE**: _Deploying the migrators as a separate cluster is optional, however the migration process is memory hungry, so you may need to devote more resources to the Orca process._
+
+### Enabling DualExecutionRepository
```
dorbin: These three headings, can you make them all imperative? 

Enable `DualExecutionRepository`
Deploy a migration cluster
(the last one is already imperative :)

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+    enabled: true
+    primaryClass: com.netflix.spinnaker.orca.sql.telemetry.SqlInstrumentedExecutionRepository
+    previousClass: com.netflix.spinnaker.orca.telemetry.RedisInstrumentedExecutionRepository
+  sql:
+    enabled: true
+  redis:
+    enabled: true
+```
+
+Note that all three execution repositories are enabled. Orca will fail to start up if the `DualExecutionRepository` is misconfigured.
```
dorbin: I see the two: sql and redis. What's the third?

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Deploying a Migration Cluster
+
+At Netflix, we deployed `orca-main`, the cluster that serves our production traffic, as well as `orca-main-sqlmigration`, which does not receive API traffic nor processes the work queue. It's sole purpose is to shovel bits from Redis to SQL.
```
dorbin: s/processes/process/

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+
+Building atop the configuration for `DualExecutionRepository`, in orca.yml:
```
dorbin: "Building atop..." 
This idiom might not be clear to non-native English speakers. 

What about just...
"Add the following configuration to `orca.yml`:

guides/operator/orca-redis-to-sql.md
```diff
@@ -0,0 +1,100 @@
...
+
+Once all executions have been migrated, you can deploy Orca once again without `DualExecutionRepository` and delete the migration cluster.
+
+## SQL-specific Metrics
+
+Orca emits a few handfuls of metrics that are specific to SQL.
+
+The `SqlHealthcheckQueueActivator` class, which will disable work queue processing if SQL connectivity goes unhealthy, emits one metric: `sql.queueActivator.invocations` with a tag of `status` (`disabled` or `enabled`).
+
+The ExecutionRepository will emit a bunch of invocation and timing metrics with the following patterns:
```
dorbin: Backticks for ExecutionRepository?

setup/productionize/persistence/orca-sql.md
```diff
@@ -0,0 +1,132 @@
...
+This guide will go over MySQL setup, how to configure Orca, as well as how to perform a zero-downtime migration from Redis to SQL.
+
+If you already have an Orca deployment, you should also refer to the [Redis to SQL Migration Guide](/guides/operator/orca-redis-to-sql/).
+
+## MySQL 5.7 Setup
+
+Orca ships with MySQL drivers by default, but you can include your own JDBC drivers on the classpath if you need to connect to a different database.
+
+Orca has been developed and tested targeting MySQL 5.7. As part of this, setting MySQL's `tx_isolation` value to `READ-COMMITTED` is essential to successfully running Orca in SQL. 
+While Orca will attempt to set this on connection sessions, it is better to have it set on the database itself.
```
dorbin, line 23: @robzienert would it be possible to change this to a procedure? 
1. Set `tx_isolation` to `READ-COMMITTED` on your MySQL.
2. Set up the schema and database uses [users?]:
   ```sql
   CREATE SCHEMA `orca`...
   etc.
```
etc.

File: setup/productionize/persistence/orca-sql.md [hunk 19]
```diff
@@ -0,0 +1,132 @@
...
+
+Before deploying Orca, the schema and database uses must first be manually setup:
+
+```sql
+CREATE SCHEMA `orca` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+```
+
+Orca is configured to support a `migration` user and a `service` user.
+The `migration` user will only be used to perform schema changes on the database, whereas the `service` user will be used for non-migration traffic.
```
dorbin: "non-migration traffic"... 
runtime traffic?

setup/productionize/persistence/orca-sql.md
```diff
@@ -0,0 +1,132 @@
...
+    redis: false
+```
+
+## Netflix's Amazon Aurora Example
+
+While vanilla MySQL provides more durability and performance over Redis, Netflix additionally uses Amazon Aurora MySQL 5.7.
+If you'd like to configure Orca to use Aurora as well, here is how Netflix has it set up.
+
+**IMPORTANT**: This configuration is for multi-region Aurora replication.
+If you are only deploying Aurora into a single region, you should not enable any binlog settings.
```
dorbin: s/you should not/don't/

dorbin (commented):
Just a few further edits/questions

guides/operator/orca-redis-to-sql.md
tillig: Might be good to point to some sort of overview guide of _where_ `orca.yml` is for the folks (like me) using Halyard to deploy. I see an `orca.yml` in both `.hal/default/staging/orca.yml` and `.hal/default/staging/registry/orca.yml`. Which one do I modify?

robzienert:
Updated and moved things around.

robzienert:
Will be merging this and will followup with other changes in subsequent PRs as necessary.

yolandabecool:
Question on the orca to sql documentation. For the setup, it mentions to set baseline for Orca to talk to SQL, in orca.yml. Should this be set in ~/.hal/default/profiles/orca-local.yml? Should we rewrite the entire file in in ~/.hal/default/profiles/orca.yml? It also seems like this only starts working on 1.10 - -should this be addressed in the documentation? 

Also, If you are not  Deploying a Migration Cluster should you stop at Enable DualExecutionRepository? Will the migration agent start to run with no further configuration? After upgrading to 1.10.1 all newly executed pipelines show up but historical data does not seems to be migrating over.

robzienert:
Ping @lwander -- I don't really know much about Halyard, can it support arbitrary configs or do we have to update halyard to explicitly support these configs?

As for the migration cluster, if you're not migrating data, then you don't need to use the DualExecutionRepository. If you're migrating just without the cluster, you will still need the DualExecutionRepository.

PR #992, opened 2018-08-27
update pipeline test guides

chinaboard:
update pipeline test guides & add api reference doc


guides/user/pipeline/expressions/index.md
```diff
@@ -208,47 +208,30 @@ the pipeline execution JSON.
     ![](images/execution-json.png)
 
 3. The execution ID is the value of the top-level **id** field in the JSON, and
-looks something like _01CGYV7Q4PMEBDYS146CCED0M6_.
+looks something like 6d2e7b36-5132-46ca-8a5a-fc4c52e6d9a6.
```
lwander: I would keep the ULID the same, we stopped using UUIDs after the 1.7 release

PR #993, opened 2018-08-28
fix(sizing): explain new sizing behavior clearly

lwander:
@jfillo fyi

### DIFF
--- a/reference/halyard/component-sizing.md
+++ b/reference/halyard/component-sizing.md
@@ -15,21 +15,26 @@ The following Kubernetes settings can be tweaked within custom sizing.
 
 ### CPU & Memory requests/limits
 
-Container requests and limits for cpu and memory can be specified in the customSizing section of the Halyard Configuration. Note that if you specify the service name, these sizings will also be applied to any sidecars running alongside the primary container. To apply sizing to only a specfic container and not the sidecars, specify the container name instead of the service name. Please see the example for details.
+Container requests and limits for cpu and memory can be specified in the `deploymentEnvironment.customSizing` section of the halconfig file. 
 
-```
+There are two ways to specify requests and limits
+
+1. _(Recommended)_ By supplying the container name, e.g. `echo:`, followed by the container requests and limits.
+
+   This sets resource configuration for the echo container in `spin-echo` service's pod only, not any sidecars
+   (e.g. the monitoring daemon).
+   
+2. By supplying the service name, e.g. `spin-clouddriver:`, followed by the container requests and limits.
+
+   This sets the resource configuration for the clouddriver container as well as any sidecar containers in the
+   `spin-clouddriver` service.
+
+Here is an example of this configuration:
+
+```yaml
 deploymentEnvironment:
   customSizing:
-    # This applies sizings to the cloudriver container as well as any sidecar 
-    # containers running with clouddriver.
-    spin-clouddriver:
-      limits:
-        cpu: 1
-        memory: 1Gi
-      requests:
-        cpu: 250m
-        memory: 512Mi
-    # This applies sizings to only the echo container and not to any sidecar 
+    # 1. This applies sizings to only the echo container and not to any sidecar 
     # containers running with echo.
     echo:
       limits:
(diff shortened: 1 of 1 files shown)


reference/halyard/component-sizing.md
```diff
@@ -15,21 +15,26 @@ The following Kubernetes settings can be tweaked within custom sizing.
 
 ### CPU & Memory requests/limits
 
-Container requests and limits for cpu and memory can be specified in the customSizing section of the Halyard Configuration. Note that if you specify the service name, these sizings will also be applied to any sidecars running alongside the primary container. To apply sizing to only a specfic container and not the sidecars, specify the container name instead of the service name. Please see the example for details.
+Container requests and limits for cpu and memory can be specified in the `deploymentEnvironment.customSizing` section of the halconfig file. 
 
-```
+There are two ways to specify requests and limits
```
dorbin, line 20: Even though you say there are "two ways," use bullets instead of numbers, because they're not meant to be done in order.

jfillo:
This reads much better, thanks!

reference/halyard/component-sizing.md
```diff
@@ -38,15 +43,31 @@ deploymentEnvironment:
...
+        cpu: 250m
+        memory: 512Mi
 ```
 
 Limits and requests follow the Kubernetes conventions [documented here](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/).
 
 #### Updating JAVA_OPTS
 
-The shaping of the pods via request and limits will not cap the java processes for the Spinnaker microservice. This can be achieved by using the `env` key in [service-settings](/reference/halyard/custom/#tweakable-service-settings).
+As of Halyard 1.8, all JVM-based services will automatically have the following JAVA_OPTS set:
```
dorbin: s/will automatically have the following JAVA_OPTS set/have the following default JAVA_OPTS/

reference/halyard/component-sizing.md
```diff
@@ -38,15 +43,31 @@ deploymentEnvironment:
...
+
+```
+JAVA_OPTS=-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MaxRAMFraction=2
+```
+
+This sets the JVMs heap to half the memory allocated per container. This can be overriden by specifying your own `JAVA_OPTS`
```
dorbin: "JVM's heap"?
or "JVM heap"?

reference/halyard/component-sizing.md
```diff
@@ -38,15 +43,31 @@ deploymentEnvironment:
...
+
+```
+JAVA_OPTS=-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MaxRAMFraction=2
+```
+
+This sets the JVMs heap to half the memory allocated per container. This can be overriden by specifying your own `JAVA_OPTS` 
+using the `env` key in [service-settings](/reference/halyard/custom/#tweakable-service-settings).
 
-In general, the `-Xms` will follow 80%-90% of the requests memory allotment and `-Xmx` will follow 80-90% of the limits memory allotment. For the clouddriver example above, the `env` key is as follows:
+In general, the `-Xms` should be 80%-90% of the requests memory allotment and `-Xmx` should be 80-90% of the limits memory allotment. For the clouddriver example above, the `env` key is as follows:
```
dorbin: Can we be more specific about the "in general" part? And "should"?

Like:
"Set -Xms to 80-90% of the requests memory allotment, unless..."

dorbin (changes requested):
Thanks for the update, Lars. 
Some minor edits.

reference/halyard/component-sizing.md
lwander: Ideally yes, but we simply don't know/have the experience.

reference/halyard/component-sizing.md
lwander: The former, fixing

lwander:
Updated

dorbin (approved):

PR #1006, opened 2018-09-07
Replace HipChat with Slack

ScottBrenner:
[HipChat is migrating to Slack](https://slackhq.com/atlassian-and-slack-partnership), might as well remove it from the documentation.


dorbin (commented):
Hi, Scott. This sounds like a good idea. 

There are a few other places in the docs that mention HipChat, can you remove it from there too?

dorbin:
@lwander do you know if we are, or need to be, doing anything about existing HipChat support, or do we expect that Slack will handle any incoming HipChat request that anyone has configured?

ScottBrenner:
@dorbin Done!

lwander:
Before we remove it -- what's the migration plan for existing HipChat users?

ScottBrenner:
Been two months, any news?

dorbin:
@ScottBrenner 
I think until @someone_in_the_community does something to change our support for HipChat, or until we know HipChat support is broken, I think we'll have to keep the current doc.

Feel free to re-open this if you determine HipChat's broken or if there's a PR on spinnaker/spinnaker changing this.

PR #1009, opened 2018-09-10
Add Wercker to navigation bar

desagar:
Added Wercker CI to the nav bar.


dorbin (approved):
Good catch! Thanks @desagar!

PR #1012, opened 2018-09-13
Fix spinnaker clouddriver role json

akrymets:
Deleted comma led to the "Bad string" error in AWS.
It was either a typo or an old syntax for IAM-roles
Deleting the comma fixes the problem.


emjburns:
LGTM thanks!

PR #1020, opened 2018-09-20
fix(codelab): Add command to target cluster to gke-rbac codelab

ezimanyi:
If a user follows the steps in this codelab in order, they'll deploy using a debian install to the machine running halyard rather than to their kubernetes cluster. Add the command that configures halyard to deploy to the kubernetes cluster.


ezimanyi:
Addresses the issue raised here: https://community.spinnaker.io/t/cant-deploy-to-google-kubernetes-engine-with-rbac/678

avh121 (approved):
Thanks!

PR #1022, opened 2018-09-26
feat(provider/kubernetes): document load balancer annotation

lwander:
https://github.com/spinnaker/spinnaker/issues/3377

### DIFF
--- a/reference/providers/kubernetes-v2/index.md
+++ b/reference/providers/kubernetes-v2/index.md
@@ -150,6 +150,26 @@ command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
   If instead Spinnaker is deploying ReplicaSets directly without a Deployment,
   this annotation does the job.
 
+## Traffic
+
+* `traffic.spinnaker.io/load-balancers`
+
+  As of Spinnaker 1.10, you can specify which load balancers
+  ([Services](https://kubernetes.io/docs/concepts/services-networking/service/))
+  a workload is attached to at deployment time. This will automatically set the
+  required labels on the workload's Pods to match that of the Services' [label
+  selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors).
+
+  This annotation must be supplied as a list of `<kind> <name>` pairs where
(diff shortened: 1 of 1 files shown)


danielpeach (approved):

reference/providers/kubernetes-v2/index.md
```diff
@@ -150,6 +150,26 @@ command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
   If instead Spinnaker is deploying ReplicaSets directly without a Deployment,
   this annotation does the job.
 
+## Traffic
+
+* `traffic.spinnaker.io/load-balancers`
+
+  In Spinnaker 1.10, you can specify which load balancers
```
dorbin: Prefer "*As of* Spinnaker 1.10" (unless you're going to remove this in 1.11+ ;)

But ideally we wouldn't even refrence the version. Just say "You can specify which load balancers..."

This is why I want to enable versioned docs.

reference/providers/kubernetes-v2/index.md
```diff
@@ -150,6 +150,26 @@ command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands
...
+  a workload is attached to at deployment time. This will automatically set the
+  required labels on the workload's Pods to match that of the Services' [label
+  selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors).
+
+  This annotation must be supplied as a list of `<kind> <name>` pairs where
+  `kind` and `name` refer to the load balancer in the same namespace as the 
+  resource. For example:
+
+  * `traffic.spinnaker.io/load-balancers: '["service my-service"]'` attaches to
+    the Service named "my-service".
```
dorbin: Backticks around the 2nd `my-service`?

dorbin (approved):
I like that in this section you provide examples rather than just the labels.

For example, under Strategy you've got `strategy.spinnaker.io/max-version-history`. But here, you show the annotation *and* its value: `traffic.spinnaker.io/load-balancers: '["service my-service"]'`. This is better.

PR #1030, opened 2018-10-02
Improve script-stage instructions

justinharringa:
* Add user token link
* No need to SSH to host
* Your user only needs Job creation access, not admin access.
* Fix curl quoting


justinharringa:
CC: @DmitryKiselev

#Hacktoberfest!

tomaslin:
LGTM

PR #1031, opened 2018-10-03
docs(ha): Add paragraph indicating that Redis endpoint must include protocol and port.

joonlim:


joonlim:
@dorbin @jtk54 @lwander

reference/halyard/high-availability.md
```diff
@@ -51,6 +51,8 @@ Although by default the three Clouddriver services will communicate with the glo
 hal config deploy ha clouddriver edit --redis-master-endpoint $REDIS_MASTER_ENDPOINT --redis-slave-endpoint $REDIS_SLAVE_ENDPOINT
 ```
 
+The values for `REDIS_MASTER_ENDPOINT` and `REDIS_SLAVE_ENDPOINT` must be a valid full Redis endpoint, not just the host name. This usually means that the endpoints should start with a `redis://` protocol and end with a `:6379` port. An example could be `redis://127.0.0.1:6379`.
```
dorbin: s/a valid full Redis enpoint/valid full Redis endpoints/

s/the hostname/hostnames/

s/An example code be/For example:/

dorbin: "...usually means...should start with..."

Are there cases we support where the endpoint won't start with `redis://` and end with that port.

dorbin (changes requested):

joonlim:
@dorbin Updated to direct to doc describing redis URI standard

dorbin (approved):

Issue #1035, opened 2018-10-03
Move 'Runbooks' to 'Operator How-Tos'


demobox:
![image](https://user-images.githubusercontent.com/223702/46434028-6a3c0780-c720-11e8-9e65-2df7b983bef0.png)

The content of 'Runbooks' is focused on the operator/admin experience.

dorbin:
#1036

PR #1036, opened 2018-10-04
docs(ops): moves Runbooks under Operator How-Tos

dorbin:


_data/navigation.yml
```diff
@@ -305,6 +305,15 @@ guides:
     children:
       - title: Custom Webhook Stages
         url: /guides/operator/custom-webhook-stages/
+      - title: Runbooks
```
demobox, line 308: Any particular reason to keep "Runbooks" as an intermediate item, rather than moving its children directly under "Operator How-Tos"?

dorbin: I put it here per your #1035. But since #988, there are now some conflicts. I'm going to close this PR and then revisit the organization of the ops guides.

PR #1039, opened 2018-10-05
fix typo in kubernetes.md

seeekr:


dorbin (approved):
Thanks for the fix, @seeekr !

PR #1045, opened 2018-10-09
Small grammar fix

CH-DanReif:


dorbin (approved):
Good catch. Thanks, @CH-DanReif

PR #1049, opened 2018-10-12
fix(provider/kubernetes): remove deployment edit from RBAC instructions

ghost:
The GKE RBAC guide seems to have gotten a little mixed up here.

From reading through the guide I don't think that `my-test-account` is
intended to run Spinnaker itself. The user is required to already be
running a Spinnaker instance somewhere, per the prerequisites.

The intention is to help a user configure a new account, named
`my-test-account`, that has RBAC configured correctly so that Spinnaker
can deploy resources into it.

This has caused a bit of confusion, namely in this issue: https://github.com/spinnaker/spinnaker/issues/3384#issuecomment-428680937


lwander (approved):

lwander:
Thanks!

PR #1050, opened 2018-10-12
feat(provider/kubernetes): simplify token creation

lwander:
The prior instructions were harder to follow by requiring several separate copy/paste commands

### DIFF
--- a/setup/install/providers/kubernetes-v2/index.md
+++ b/setup/install/providers/kubernetes-v2/index.md
@@ -52,41 +52,16 @@ even when managing multiple Kubernetes clusters. This can be useful if you need
 to grant Spinnaker certain roles in the cluster later on, or you typically
 depend on an authentication mechanism that doesn't work in all environments.
 
-Given that you want to create a Service Account in context `$CONTEXT`, create
-the following resources with `kubectl apply --context $CONTEXT -f
-https://spinnaker.io/downloads/kubernetes/service-account.yml` 
-
-```yaml
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: spinnaker-service-account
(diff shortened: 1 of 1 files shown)


ghost (approved):

setup/install/providers/kubernetes-v2/index.md
```diff
@@ -52,65 +52,29 @@ even when managing multiple Kubernetes clusters. This can be useful if you need
...
-> example. For more fine-grained control, see the
-> [RBAC](#optional-configure-kubernetes-roles-rbac) section.
-
-Next, grab the token that this service account relies on:
+Given that you want to create a Service Account in existing context `$CONTEXT`,
+the following commands will create `spinnaker-service-account`, and add its
+token under a new user called `${CONTEXT}-token-user` in context `$CONTEXT`.
 
 ```bash
+# This service account uses the ClusterAdmin role -- this is not necessary,
```
dorbin, line 60: @lwander
Does this live on as a Bash script, other than to be copied from here? Reason I ask is the comment at the beginning confuses me, and I wonder if it should be moved up as additional explanation before you show the commands. And maybe include more explanation about the "not necessary" part.

dorbin (commented):
One question, otherwise LGTM.

setup/install/providers/kubernetes-v2/index.md
lwander: Yes -- the command includes a link to account details. I added the comment here to make clear where the account is created since it includes all details in the link in the following line.

dorbin: 👍

dorbin (approved):

PR #1052, opened 2018-10-16
fix(docs): yaml to yml

vuza:
Changed not working service-settings/*.yaml to working service-settings/*.yml


vuza:
This change is important because if you follow the current docs you get `Requested tenant identifier '${azuretenantid}' is not valid.`

lwander (approved):
Good catch!

PR #1053, opened 2018-10-17
docs(provider/kubernetes): document --only-spinnaker-managed flag

mnalsup:
Added documentation to the kubernetes account add/edit commands to set the only-spinnaker-managed flag. Related to https://github.com/spinnaker/halyard/pull/1067.


dorbin:
@mnalsup This page is generated. Doc for --only-spinnaker-managed should be added automatically.

PR #1055, opened 2018-10-17
docs(kubernetes): update docs on overriding artifacts when patching

maggieneterval:


lwander (approved):
Thanks! totally forgot about the guide as well, thanks for catching that

PR #1059, opened 2018-10-19
Fix typo: s/application/pipeline/

jeremydonahue:
This fixes a typo in one of the cli examples. `spin application list` should be `spin pipeline list` in this case.


jtk54 (approved):
Thanks for the fix.

PR #1067, opened 2018-10-23
docs(ha): Add section explaining how to delete orphan services.

joonlim:


joonlim:
@dorbin @lwander @jtk54

reference/halyard/high-availability.md
```diff
@@ -135,6 +135,14 @@ This service's name when [configuring its sizing](/reference/halyard/component-s
 
 To add a [custom profile](/reference/halyard/custom/#custom-profiles) or [custom service settings](/reference/halyard/custom/#custom-service-settings) for this service, use the name `echo-worker`.
 
+## Deleting Orphan Services
```
lwander: I would call these "orphaned" services

dorbin: including in the option `--delete-orphaned-services`?

reference/halyard/high-availability.md
```diff
@@ -135,6 +135,14 @@ This service's name when [configuring its sizing](/reference/halyard/component-s
...
+
+When enabling or disabling HA for a service on a running Spinnaker, Halyard will not clean up the old service(s) by default. This means that if a non-HA Clouddriver is running and Spinnaker is then deployed with HA Clouddriver enabled, the non-HA Clouddriver will still be running, even though it is no longer used. To clean up these orphan services, add a `--delete-orphan-services` flag to `hal deploy apply`:
```
jtk54: Update here if you change the flag name.

jtk54 (approved):

reference/halyard/high-availability.md
dorbin: But not just Clouddriver, right? Can be any service? If so, I would add a "for example" to the above:

"This means that if a non-HA Clouddriver is running (for example)  and Spinnaker is then deployed..."

dorbin: The first sentence seems to  contradict the rest of the description. Mostly that's probably my lack of understanding, but if Halyard is going to clean up old services, I would not expect the non-HA Clouddriver to be still running. Could you clarify?

lwander (approved):

reference/halyard/high-availability.md
dorbin: nm, I read "Halyard will clean up" rather than "Halyard will not clean up."

Carry on.

dorbin (approved):

PR #1069, opened 2018-10-23
correct redis URI for database selection

rchench:
the original URI (redis://admin:passw0rd@some.redis.url:6379?db=1) does not work. it always selects db 0.
it has to be changed to redis://admin:passw0rd@some.redis.url:6379/1, then it works.
I verified this in redis 3.x and 4.x. not sure about redis 2.x. but I guess no one is using redis 2.x for production environment. so I simply suggest to update the URI.


duftler (approved):
LGTM.

PR #1079, opened 2018-10-29
docs(spin): Adds instructions for google out of band oauth2 token.

jtk54:


guides/spin/cli/index.md
```diff
@@ -90,14 +92,47 @@ auth:
...
+  oauth2:
+    authUrl: # OAuth2 provider auth url
+    tokenUrl: # OAuth2 provider token url
+    # no clientId or clientSecret
+    scopes: # Scopes requested for the token
+    - scope1
+    - scope2
+    cachedToken:
+      accesstoken: ACCESS_TOKEN # Note the key capitalization
+      refreshtoken REFRESH_TOKEN # Note the key capitalization
```
lwander: Missing a `:`

guides/spin/cli/index.md
```diff
@@ -90,14 +92,47 @@ auth:
...
+  enabled: true
+  oauth2:
+    authUrl: # OAuth2 provider auth url
+    tokenUrl: # OAuth2 provider token url
+    # no clientId or clientSecret
+    scopes: # Scopes requested for the token
+    - scope1
+    - scope2
+    cachedToken:
+      accesstoken: ACCESS_TOKEN # Note the key capitalization
```
lwander: What needs to be capitalized here? I see ACCESS_TOKEN as a placeholder? If it is, you should wrap in `${}` or similar to make it obvious.

lwander: Also, I noticed that `cachedToken` is camelCase, but `accesstoken` isn't...

guides/spin/cli/index.md
```diff
@@ -90,14 +92,47 @@ auth:
...
+    - scope2
+    cachedToken:
+      accesstoken: ${ACCESS_TOKEN} # Note the key capitalization
+      refreshtoken: ${REFRESH_TOKEN} # Note the key capitalization
+```
+
+This method is OAuth2-provider specific since the workflow to acquire
+a token is different for each provider.
+
+* Google OAuth2 with `gcloud`
```
dorbin: One bullet by itself is kinda lonely. Maybe make this a second sentence in the prev. paragraph? 
"To do so for Google OAuth2:
1. Authenticate via `gcloud auth login`
2. Use the following commands to acquire the tokens:
  - (commands)
"

dorbin (approved):
Small suggestion; otherwise LGTM

guides/spin/cli/index.md
jtk54: The idea is that we'd have a list for all the supported providers. Ideally this would be its own header, but markdown headers don't nest 5 levels deep unfortunately. Should I still roll this into the previous paragraph?

dorbin: What do you think is the timeframe for having n+1?

jtk54: I don't know, we'll address this when we have to.

PR #1080, opened 2018-10-31
Add Cloud Foundry to Reference / Providers nav; add 'alpha' note for CF

fifthposition:


reference/providers/cf.md
```diff
@@ -9,6 +9,8 @@ sidebar:
 
 If you are not familiar with Cloud Foundry (CF) or any of the terms used below, please consult [Cloud Foundry's reference documentation](https://docs.cloudfoundry.org).
 
+**NOTE:** The Cloud Foundry provider is currently in alpha phase.
```
dorbin: @lwander 
Do you think this should use the alpha notice format you used in places like [this](https://www.spinnaker.io/guides/user/kubernetes-v2/patch-manifest/)?

lwander: Oh yes that'd be great.

dorbin: Cool, @fifthposition can you replace with this:
`{% include alpha version="1.10" %}`

Probably version 1.10, right?

fifthposition:
@dorbin @lwander Done! Many thanks. :)

dorbin (approved):
Thanks, Ben!

PR #1115, opened 2018-11-22
docs(setup): Updated hal command to start individual services

scheibinger:
There is no option like `--service-name` in the latest halyard binary


dorbin (approved):
Looks good. Thanks @scheibinger

PR #1117, opened 2018-11-28
Update quick-install.yml file for K8s to use 1.9.5

MrTrustor:
I will test this file with 1.10/1.11 in the near future.
In the meantime, we can safely go from 1.9.0 to 1.9.5.


duftler (approved):
LGTM.

PR #1141, opened 2018-12-09
docs(setup): add details for configuring ECR

jaronsummers:
This setup requires passwordCommand and that AWS be baked into the clouddriver image, which is not available on a stable release, but is working correctly in `master-latest-unvalidated`.


ethanfrogers (approved):

ethanfrogers:
@edwinavalos can you take a look at this as well?

edwinavalos (approved):
Looks good, thanks for writing these up.

PR #1148, opened 2018-12-14
docs(caching/externalize-redis): Fixed filepath for $SERVICE-local.yml

jlumme:
Filepath was pointing to 
~/.hal/$DEPLOYMENT/profile-settings/$SERVICE-local.yml
while the correct path is
~/.hal/$DEPLOYMENT/profiles/$SERVICE-local.yml


lwander (approved):
Thank you!

PR #1150, opened 2018-12-16
Make $CONTEXT explicit

sameg14:
Reduce cognitive overhead for new users trying to figure out what `$CONTEXT` is and where it comes from


lwander (approved):
:+1: in favor of this

PR #1162, opened 2018-12-20
docs(spin): Adds spin pipeline templates docs.

jtk54:


dorbin (approved):
Is there also a `spin pt`, for pipeline templates? I thought that's what I was seeing this mornning in your Medium post.

Also, I think eventually this should go in /reference/. 
A `spin` guide should be more like...

## Delete a pipeline
`spin pi delete`

...etc.

jtk54:
Agreed on the /reference reorganization, though I think I'll save that for when I automate the spin CLI documentation.

PR #1163, opened 2018-12-21
fix(codelab): Update codelab to change Jenkins port

ezimanyi:
When following the Hello Deployment codelab, front50 doesn't start up as its port conflicts with the default port for Jenkins. Add instructions to change the Jenkins port to avoid this issue.


ezimanyi:
Address https://github.com/spinnaker/spinnaker/issues/3780

dorbin (approved):

PR #1171, opened 2019-01-04
Fix Cluster Role Binding for Pod Logs

oliverisaac:
The cluster role binding for pod logs should be "pods/log" singular not "pods/logs" plural.

https://kubernetes.io/docs/reference/access-authn-authz/rbac/


lwander (approved):
I wonder if this changed across releases of k8s

PR #1175, opened 2019-01-11
feat(reference/architecture): add halyard to the dependancy graph

kevinawoo:
# Before/After:
<kbd>ezgif com-gif-maker</kbd>


ttomsu:
The coloring implies that Halyard is completely outside of Spinnaker, which it isn't, nor is it a runtime dependency. I suggest adding a third color for CLI + daemon and update the text.

kevinawoo:
# Here's what it looks like now:
<kdb></kdb>

bpowell:
LGTM thanks @kevinawoo

ttomsu (approved):

PR #1177, opened 2019-01-22
feat(stages): add note that manual judgement requires echo

ghost:
Here's a recent issue that stems from this not being clear in a stacktrace and seemingly not documented anywhere (at least nowhere I can find on the first page of google results): https://github.com/spinnaker/spinnaker/issues/3860


dorbin (approved):
👍

PR #1186, opened 2019-01-23
Note how to verify redis changes worked

lwander:


dorbin (commented):
Might work better to say...

"
To confirm this works:
1. Run `hal config generate`.
1. Run `hal deploy apply` 
(etc.)

lwander:
I don't want to make `hal deploy apply` a part of the "verify this works", since ideally the verification shouldn't require making changes to live infra.

lwander:
@dorbin updated

dorbin (approved):
:+1:

PR #1188, opened 2019-01-24
Update commands.md with a default value

davidxia:
for `--check-permissions-on-startup`.


davidxia:
cc @lwander

lwander:
These docs are auto-generated, and the auto generator ignores `null` (meaning there is no default). We could update that behavior, but I think it's been OK so far.

PR #1215, opened 2019-02-11
feat(core): Add Halyard Maven Artifact account support

jmelchio:
- spinnaker/spinnaker#3954
- spinnaker/spinnaker#3942


dorbin:
Same comment as I put in your other PR just now, @jmelchio . The hal command reference is generated.

PR #1218, opened 2019-02-12
feat(cf): update setup docs to match Halyard additions

jmelchio:
- spinnaker/spinnaker#3961


dorbin:
@jkschneider can you take a look? Thanks.

PR #1221, opened 2019-02-13
Add administrator functionality documentation

PerGon:
Feel free to hold merging as halyard configuration is not done (neither implemented) yet. But provide feedback nonetheless.

I'll be implementing this config in halyard soon and then I'll update the docs accordingly.

### DIFF
--- a/setup/security/admin/index.md
+++ b/setup/security/admin/index.md
@@ -1,0 +1,32 @@
+---
+layout: single
+title:  "Administrator functionality"
+sidebar:
+  nav: setup
+---
+
+{% include toc %}
+
+## Introduction
+In Spinnaker, it is possible to define that users belonging to a certain role are considered "Administrators". This virtually removes all READ/WRITE restrictions to accounts and applications for these users.
+
+{% include
+   warning
+   content="This feature gives God Mode like capabilities to the users who are admins. Proceed with caution."
+%}
+
+## Enable and Configure Admin functionality
+
+### Halyard
+TBD
+
+### Manually add configuration in fiat
+
+In the fiat config file, add the following:
+
+```yaml
+fiat:
+  admin:
+    roles:
+      - "role of the spinnaker gods"
+```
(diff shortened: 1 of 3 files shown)


ethanfrogers:
@PerGon I'd think this is okay to merge before the functionality is built into Halyard. At the very least it communicates that the feature is available which is a step above what we had previously.

setup/security/admin/index.md
```diff
@@ -0,0 +1,32 @@
...
+   warning
+   content="This feature gives God Mode like capabilities to the users who are admins. Proceed with caution."
+%}
+
+## Enable and Configure Admin functionality
+
+### Halyard
+TBD
+
+### Manually add configuration in fiat
```
ethanfrogers: Small nit, capitalize Fiat.

setup/security/admin/index.md
```diff
@@ -0,0 +1,32 @@
...
+
+In the fiat config file, add the following:
```
ethanfrogers: Same capitalization fix as above. Also, if we want to merge this before the Halyard work is done you might include a path to the file where this should be added (`~/.hal/$DEPLOYMENT/profiles/fiat-local.yml`).

setup/security/admin/index.md
```diff
@@ -0,0 +1,32 @@
...
+
+```yaml
+fiat:
+  admin:
+    roles:
+      - "role of the spinnaker gods"
```
ethanfrogers: It might be more useful to have an example of what should actually go here. Something like:
```
- devops-admin
```

ethanfrogers (approved):
LGTM, just a couple of suggestions!

setup/security/admin/index.md
kuberkaul: 👍

setup/security/admin/index.md
kuberkaul: maybe a note on if this using saml service (okta) then these are okta groups and you would need to run `hal deploy apply` after the change.

kuberkaul:
tested the feature and it works pretty well, thanks @PerGon  !

PerGon:
Thanks for the feedback :)
Applied suggestions

PR #1228, opened 2019-02-21
PubSub messages need to be valid JSON

milesmatthias:
When doing some testing, I noticed that if my pubsub messages aren't JSON, an exception is raised in echo, which fails to trigger my pipeline. This line would've been helpful.


setup/triggers/google.md
```diff
@@ -63,6 +63,9 @@ At this point, you can use the [publisher
 guide](https://cloud.google.com/pubsub/docs/publisher){:target="\_blank"} to
 learn how to publish messages programmatically to the topic you have created.
 
+Note that your topic messages need to be valid JSON, otherwise an exception
```
dorbin, line 66: @milesmatthias 
Can you make this....

> **Note**: your topic message must be valid JSON.

I don't think we need the phrase about the exception. But as the user who encountered this, if you think it's helpful feel free to leave it in.

dorbin (changes requested):

setup/triggers/google.md
milesmatthias: Well I was doing testing to see if the topic message was getting to Spinnaker, so the fact that there was an exception in echo was telling me it was getting there, just that the message was malformed. So I'd probably like to add `in the echo service` at the end of that sentence.

dorbin (approved):

PR #1236, opened 2019-03-01
feat(sigs): add SIG meeting calendar, working slack links

erikmunson:
Right now it's a bit difficult for community members to keep track of when each SIG meeting takes place and/or notice when new SIG meetings start up. Got some positive feedback on the idea of creating a public calendar that folks could subscribe to which included all the various SIG meetings.

The calendar is owned by `sig@spinnaker.io`, and I've included a couple ways to subscribe to it along with an embedded view of upcoming events. We've got the as-code SIG partially on there now but I'll be reaching out to the SIG leads over the next few days to get their events onto the calendar as well.

While I was in here I also replaced the slack channel links (which just pointed to the overall workspace) and made them standard Slack deep links to the appropriate channel.

cc @bpowell @ethanfrogers @robfletcher @emjburns @helenelau

### DIFF
--- a/community/governance/sigs.md
+++ b/community/governance/sigs.md
@@ -18,8 +18,22 @@ Below is a list of the current Spinnaker SIGs.
 
 | Name | Lead(s) | Agenda/Notes | Meeting Link | Contact | Schedule |
 |-|-|-|-|-| - |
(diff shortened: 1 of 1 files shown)


ethanfrogers (approved):
+1 +1 +1

emjburns:
LGTM

PR #1253, opened 2019-03-13
Update Gateway troubleshooting FAQ

norajam:
Update one more FAQ for gateway service can't be connected

### DIFF
--- a/setup/quickstart/faq.md
+++ b/setup/quickstart/faq.md
@@ -31,7 +31,7 @@ following issues:
    These settings will configure cross-origin resource sharing (CORS) between your Gate and Deck
    endpoints; if this is not properly configured, your browser will reject requests from Deck to
    Gate.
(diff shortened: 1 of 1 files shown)


ezimanyi (commented):
Thanks for the addition to the FAQ!

When redis is unavailable, does the error manifest itself with the exact error message noted in this FAQ?  If not, I'd suggest moving this into a separate question to keep this question focused just on resolving that particular error.

Also, suggested a few minor edits to the wording:
If you have a local deployment of Spinnaker, ensure that Redis is available at the configured address (`localhost:6379` by default).  If not, start redis by running `sudo systemctl enable redis-server && sudo systemctl start redis-server` and restart spinnaker by running `sudo systemctl restart spinnaker`.

norajam:
Hi Eric,

Good suggestion for the words!

The error happens without a clear message, it's linked to this exact message. 

The root cause is in local deployment, redis-server is used as the cache service for those Spinnaker micro components. However, users only see above exact message when they access gateway.

ezimanyi:
@luhaixun : Sounds good, then I think it makes sense to leave it in this question.  So if you just add a commit to update the wording per above, I can merge this.  Thanks!

norajam:
@ezimanyi I add a new commit, please let me if any questions.

ezimanyi:
This looks great, thanks!

ezimanyi (approved):

PR #1254, opened 2019-03-13
Prometheus update script

norajam:
Update to use install.sh because install is no longer valid


duftler (approved):
LGTM.

duftler:
fyi @ewiseblatt

PR #1282, opened 2019-03-26
docs(artifacts): Update docs for artifact inflation

ezimanyi:
As of Spinnaker 1.13, custom jinja templates need to be configured in igor.local.yml.

### DIFF
--- a/reference/artifacts/from-build-triggers.md
+++ b/reference/artifacts/from-build-triggers.md
@@ -33,8 +33,8 @@ messageFormat=JAR
 ```
 
 The recommended way to configure artifact templates is by using the `hal config artifact templates`
- [Halyard command](reference/halyard/commands/#hal-config-artifact-templates):
- ```
+ [Halyard command](/reference/halyard/commands/#hal-config-artifact-templates):
+```
 hal config artifact templates add <name of template> --template-path <path to the template> 
 ```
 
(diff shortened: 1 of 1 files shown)


reference/artifacts/from-build-triggers.md
```diff
@@ -32,14 +32,23 @@ properties from your CI job:
...
+
+As an alternative, you can manually configure templates by adding the following to `igor-local.yml`:
 ```yaml
 artifacts:
   templates:
   - name: <name of template>
     templatePath: <path to the template>
 ```
-You can then use the custom template by exporting the following as properties from your
+(Prior to Spinnaker 1.13, these templates needed to be confgiured in `echo-local.yml`; as of
```
dorbin: @ezimanyi 
This sentence applies only if I'm using igor-local, right? If I use the above hal command, it goes in hal config? 
The reason I ask is that this sentence is pretty strong in saying "...must be configured in igor-local.yml," and it sounds like that's true regardless of which way I configure it. 

Assuming the hal command puts it in hal config, I would say something like...
"Before Spinnaker 1.13, this fragment went into `echo-local.yml`. As of 1.13, it goes in `igor-local.yml`, or you can use hal config artifact..."

ezimanyi: @dorbin : Updated the wording, and fixed the broken link.  Let me know if this sounds better!

dorbin: :+1:

dorbin (approved):

PR #1302, opened 2019-04-04
Docs(nav): fixes obsolete link from left-nav...

dorbin:
to new GCE Source to Prod codelab.


dorbin:
In #1241 I pointed the link to the new codelab, but did not update the left-nav at that time.

PR #1311, opened 2019-04-09
feat(cf): Docs for using SpEL to retrieve service key

stuart-pollock:
spinnaker/spinnaker#4260


dorbin (commented):
LGTM, pending  review of the linked PR in spinnaker/spinnaker

claymccoy (approved):

PR #1312, opened 2019-04-10
docs(provider/azure): Update the outdated content of Spinnaker for Azure

mybayern1974:
@dorbin, could you please help take a look?


setup/install/providers/azure.md
```diff
@@ -52,16 +52,22 @@ RESOURCE_GROUP="Spinnaker"
...
 ```bash
 VAULT_NAME=<Insert Vault Name>
 az keyvault create --enabled-for-template-deployment true --resource-group $RESOURCE_GROUP --name $VAULT_NAME
 az keyvault set-policy --secret-permissions get --name $VAULT_NAME --spn $APP_ID
 az keyvault secret set --name VMUsername --vault-name $VAULT_NAME --value <Insert default username>
-az keyvault secret set --name VMPassword --vault-name $VAULT_NAME --value <Insert default password>
+az keyvault secret set --name VMSshPublicKey --vault-name $VAULT_NAME --value <Insert default SSH public key>
 ```
 
+> NOTE: If you prefer to use password instead of SSH public key, then replace<br/>
```
dorbin: Prefer Markdown over HTML. I suggest;

> NOTE: If you prefer...then replace...

```bash
az keyvault secret set --name VMSshPublicKey --vault-name $VAULT_NAME --value` `<Insert default SSH public key>
```
with...

```bash
az keyvault secret set --name VMPassword --vault-name $VAULT_NAME --value` `<Insert default password>
```

Azure has requirements...

setup/install/providers/azure.md
```diff
@@ -52,16 +52,22 @@ RESOURCE_GROUP="Spinnaker"
...
+`az keyvault secret set --name VMSshPublicKey --vault-name $VAULT_NAME --value` `<Insert default SSH public key>`<br/>
+with<br/>
+`az keyvault secret set --name VMPassword --vault-name $VAULT_NAME --value` `<Insert default password>`<br/>
+Azure has requirements for setting the username and password to provision VM. Follow the [rule](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/faq#what-are-the-username-requirements-when-creating-a-vm){:target="\_blank"} otherwise the provisioning of VM scale set will fail.
```
dorbin: This whole password section might be a little clearer if instead of starting with the `>NOTE`, you create a level-3 heading: "Using a password instead of an SSh key"

That way, these commands aren't part of a `>NOTE:` section.

dorbin: I think you can change this whole paragraph to...
"Follow the password rules documented [here](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/faq#what-are-the-username-requirements-when-creating-a-vm)"

That is, I think it's safe to assume the reader knows the rules apply to provisioning the VM and that it won't work if the password doesn't comply.

setup/install/providers/azure.md
```diff
@@ -79,11 +85,19 @@ hal config provider azure account add my-azure-account \
   --subscription-id $SUBSCRIPTION_ID \
   --default-key-vault $VAULT_NAME \
   --default-resource-group $RESOURCE_GROUP \
+  --packer-resource-group $RESOURCE_GROUP \
```
dorbin, line 92: Do we assume the user is using Packer? The reference doc for this command says "The resource group to use if baking images with Packer."

The doc also has " `` packer-storage-account`: The storage account to use if baking images with Packer." So, if we assume the user is using Packer, should we include this too?

setup/install/providers/azure.md
```diff
@@ -79,11 +85,19 @@ hal config provider azure account add my-azure-account \
...
   --app-key
 ```
 
-> NOTE: You will be prompted for the App Key on standard input. If necessary,
-you can generate a new key: `az ad sp credential reset --name $APP_ID`
+> NOTE:
+> 1. You will be prompted for the App Key on standard input. If necessary,
+you can generate a new key: `az ad sp credential reset --name $APP_ID`<br/>
```
dorbin: Prefer using Markdown blank lines instead of HTML line breaks.

setup/install/providers/azure.md
```diff
@@ -79,11 +85,19 @@ hal config provider azure account add my-azure-account \
...
+> 2. Starting from Halyard v1.19, SSH public key will be used to provision VM scale set by default. If you prefer to use password, add the parameter<br/>
+`--useSshPublicKey "false" \`<br/>
+Prior to Halyard v1.19, the default credential option is password and halyard does not support the above parameter to switch the option to SSH public key.<br/>
```
dorbin: If before 1.19 Halyard doesn't *support* the SSH public key option, is it correct to say that password is the *default* credential option?

dorbin (changes requested):
@mybayern1974 
Do you know of anyone else you can tag for a review who knows Azure setup well enough? I'd like to get another pair of informed eyes on this.

setup/install/providers/azure.md
mybayern1974: fixed

setup/install/providers/azure.md
mybayern1974: Fixed with not using ">NOTE". While I did not use a level-3 heading since there are no level-3 headings at other places in this doc, which means I need to reconstruct the doc a little bit to make it a "level-3 heading style". Will you agree I save the effort to do that?

mybayern1974: Fixed

setup/install/providers/azure.md
mybayern1974: Good catch! Firstly we do assume the user is using Packer. Then, the parameter "--packer-storage-account" is out of date, which means it is not expected to appear in the ref doc any more and only the parameter "--packer-resource-group" is enough. I'll initial another PR to Halyard to fix this.

setup/install/providers/azure.md
mybayern1974: Fixed

setup/install/providers/azure.md
mybayern1974: Good catch! The justification is:
1. Prior to 1.19, the default behavior is password, but there is an option to enable ssh public key at the cloud driver side with some manual config, though hal does not support set that.
2. As for users that are stick to prior-hal-v.1.19 and need the SSH key solution, we have an offline document for instruction. Indeed, we have offline sent that doc to some customers that belong to this cohort;
3. I personally do not prefer to put the content of the doc mentioned at #-2 here given this is an official doc and that instruction bypassing hal is a workaround. On the other hand, I assume things could become better once hal v1.19 comes into being in the near future;
4. I do not prefer to make a solid statement here say hal prior-v1.19 does not support SSH key, because we do support that and that statement may keep some of our potential users away. 

With these, feel free to suggest any proposals and I'm willing to follow.

setup/install/providers/azure.md
dorbin: :+1:

setup/install/providers/azure.md
dorbin: @mybayern1974
I think you can add h3s without any problems. But yeah, I think you're ok with just removing the note.

dorbin: :+1:

setup/install/providers/azure.md
dorbin: Works for me

dorbin:
@mybayern1974 
I still would like another Azure setup expert to take a look. Can you suggest anyone?

setup/install/providers/azure.md
```diff
@@ -52,15 +52,25 @@ RESOURCE_GROUP="Spinnaker"
 az group create --name $RESOURCE_GROUP --location <Insert Location>
 ```
 
-Finally, create a Key Vault (where the vault name is globally unique) and add a default username/password:
+Finally, create a Key Vault (where the vault name is globally unique) and add a default username/ssh public key. This credential is used to provision Azure VM scale set. You can log on to VM instances in the VM scale set with this credential.
```
grayzu: Should up date password on line 19 above to ssh public key as well.

grayzu: Also the credential is used for all VM scale sets by default. Correct? If so, this should be clear here.

setup/install/providers/azure.md
```diff
@@ -79,11 +89,19 @@ hal config provider azure account add my-azure-account \
...
 
-> NOTE: You will be prompted for the App Key on standard input. If necessary,
+> NOTE:
+> 1. You will be prompted for the App Key on standard input. If necessary,
 you can generate a new key: `az ad sp credential reset --name $APP_ID`
+> 2. Starting from Halyard v1.19, SSH public key will be used to provision VM scale set by default. If you prefer to use password, add the parameter\
+`--useSshPublicKey "false" \`\
+Prior to Halyard v1.19, the default credential option is password and halyard does not support the above parameter to switch the option to SSH public key.\
+Secret values of either the SSH public key or password are stored in the Azure key vault specified by the parameter `--default-key-vault`, in which the stored secret names are called "VMPassword" and "VMSshPublicKey", separately.
+> 3. The used Azure region by default is "eastus" and "westus". If you would like to add custom regions, add the parameter\
```
grayzu: Wording is a little funky. Should read "The Azure region used by default are " ...

grayzu (changes requested):
Looks great. Few minor comments added.

setup/install/providers/azure.md
mybayern1974: Good catch. Resolved

mybayern1974: > Should up date password on line 19 above to ssh public key as well.

Good catch. Resolved

mybayern1974: > Also the credential is used for all VM scale sets by default. Correct? If so, this should be clear here.

resolved

setup/install/providers/azure.md
mybayern1974: > Wording is a little funky. Should read "The Azure region used by default are " ...

resolved

dorbin (approved):

dorbin:
Thanks, @mybayern1974 and @nanxu and @grayzu

Issue #1277, opened 2019-03-25
What is Halyard?


aviv-ebates:
Page: [/setup/install/halyard/](https://www.spinnaker.io/setup/install/halyard/)

Feedback:

It's not clear what Halyard **is**, except that it has something to do with installing/managing Spinnaker.

Is it the cli I install on my personal machine, like `kubectl` and `gcloud` and `knife`, or is it a service that's part of the deployment like `logstash` and `chef` and the k8s master?

dorbin:
@aviv-ebates it's the Spinnaker-deployment-management CLI. 
#1278

aviv-ebates:
so, it's installed on my laptop, like `kubectl`?
The docs should explicitly say that. Assume whoever is reading the docs, especially the Getting Started docs, isn't familiar with the language and workflows used by the project/team.

dorbin:
It *can* be installed on your laptop. It can also be installed on a VM. Or in a [container](https://www.spinnaker.io/setup/install/halyard/#install-halyard-on-docker). 

But, I hear ya. I'll create another PR to make that clearer.

dorbin:
#1280

PR #1318, opened 2019-04-15
replace `Spinnaker` with `Halyard` where appropriate

aviv-ebates:
From my understanding of #1277, this is a more accurate description of what's going on (i.e., Halyard, the UI part of Spinnaker, is using `kubectl`, not Spinnaker itself, which is only installed on the cluster itself).


dorbin:
@aviv-ebates , that's not quite what #1277 says. The result of that issue is to clarify Halyard as *analogous* to kubectl. 

Also, Halyard isn't *the* UI part of Spinnaker. It's the CLI to administer Spinnaker, but Deck is more accurately considered the Spinnaker UI. 

Spinnaker, in general, relies on `kubectl`. It might be true that Halyard does too, but Spinnaker as a whole does.

PR #1324, opened 2019-04-19
trying to make it clearer what it means to use the namespace option

milesmatthias:
…and what you will see if you don't have this configured correctly.

See https://github.com/spinnaker/spinnaker/issues/3666#issuecomment-485001361 and https://github.com/spinnaker/spinnaker/issues/4319.

cc @ezimanyi


ezimanyi (commented):
Thanks for adding this!  I actually think the way you worded it in the Github comment is much clearer, so maybe some adaptation of that would be good in the docs.  Something like:

If you're using RBAC to restrict the spinnaker service account to a particular namespace, you must specify that namespace when you add the account to Spinnaker.  If you don't specify any namespaces, then Spinnaker will attempt to list all namespaces, which is requires a cluster-wide role, and will fail.

With that detail it probably isn't even necessary to link back to the issue itself.  Thanks!

milesmatthias:
Updated per @ezimanyi's comment above. I do think specifying the behavior you'll see if you don't have it configured correctly is helpful, since the error behavior isn't obvious as to what the issue is. (Hopefully that will be addressed through https://github.com/spinnaker/spinnaker/issues/4319)

ezimanyi (approved):
Looks good, thanks!

milesmatthias:
Thanks for all your help and merging this @ezimanyi!

PR #1328, opened 2019-04-24
docs(kubernetes): kubernetes traffic updates

maggieneterval:
First pass at docs for the new Kubernetes rollout strategies. Also cleans up some old Kubernetes docs, removing the alpha warning from the Patch Manifest and Traffic docs, and updating the Deploy Manifest images to reflect the current state of the UI.


guides/user/kubernetes-v2/rollout-strategies/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+{% include toc %}
+
+This guide describes how to take advantage of the
+[Kubernetes V2](/setup/install/providers/kubernetes-v2) provider's first-class support
+for common rollout strategies, including dark, highlander, and red/black rollouts.
+
+The implementation of these rollout strategies currently leverages Spinnaker's existing
+[traffic management strategy](/guides/user/kubernetes-v2/traffic-management/) and so is
+[valid for ReplicaSets only](/guides/user/kubernetes-v2/traffic-management/#you-must-use-replica-sets).
```
ethanfrogers, line 18: I'm not sure if we mention this elsewhere but it would be great if we could call out exactly why advanced traffic management isn't possible with Deployments. It seems like it might be a fairly common question.

guides/user/kubernetes-v2/rollout-strategies/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+{%
+  include
+  figure
+  image_path="./highlander.png"
+%}
+
+### Red/Black Rollouts
+
+Use a highlander rollout to deploy a new version of your application alongside the existing
```
ethanfrogers: s/highlander/RedBlack/

ethanfrogers (approved):
LGTM, just a couple of small things. This is YUGE!

guides/user/kubernetes-v2/rollout-strategies/index.md
```diff
@@ -0,0 +1,100 @@
...
+  nav: guides
+---
+
+{% include alpha version="1.14" %}
+
+{% include toc %}
+
+This guide describes how to take advantage of the
+[Kubernetes V2](/setup/install/providers/kubernetes-v2) provider's first-class support
+for common rollout strategies, including dark, highlander, and red/black rollouts.
```
dorbin, line 14: In /concepts/ we have a list of deployment strategies. Dark is not listed there. Seems like it would be worthwhile adding it [there](https://www.spinnaker.io/concepts/#deployment-strategies), though it would require updating that graphic. I'm not sure if we have access to the image source for that.

guides/user/kubernetes-v2/rollout-strategies/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+The implementation of these rollout strategies currently leverages Spinnaker's existing
+[traffic management strategy](/guides/user/kubernetes-v2/traffic-management/) and so is
+[valid for ReplicaSets only](/guides/user/kubernetes-v2/traffic-management/#you-must-use-replica-sets).
+
+## Rollout Strategy Options
```
dorbin, line 20: [Here](https://www.spinnaker.io/concepts/#deployment-strategies) and elsewhere we call them "deployment strategies." I think we should be consistent, if possible, but I guess that would mean a change in Deck too?

guides/user/kubernetes-v2/rollout-strategies/index.md
dorbin: Of course, Highlander's not there either. :/

guides/user/kubernetes-v2/rollout-strategies/index.md
dorbin: Then s/RedBlack/red\/black/ 
😉

dorbin (approved):
Thanks for the doc, Maggie!

guides/user/kubernetes-v2/rollout-strategies/index.md
ethanfrogers: Techincally `s/RedBlack/Red\/Black/`...right? ;)

dorbin: :laughing: 
You mean my unescaped slash? I escaped it, but didn't use backticks, so it rendered like that.

Or do you mean the caps? On [this page](https://www.spinnaker.io/concepts/#deployment-strategies), it's all lower case, except for init-caps on the image labels. I'm in favor of lower case.

guides/user/kubernetes-v2/rollout-strategies/index.md
maggieneterval: Curious for @ethanfrogers' opinion on this, but @ezimanyi and I were afraid "deployment" was too loaded a term in the Kubernetes community since it refers to a specific Kubernetes kind (a kind you can't even use with these strategies, to boot!). But if we don't think that's a huge problem I'm all for consistency and happy to update Deck's wording as well.

dorbin: Ah. Interesting. 

One possibility is that "deployment strategy" is a term unto itself, separate from, and unconfused with, "deployment." But that might not be the case for users immersed in K8s. 

I defer to y'all, Maggie and Ethan and Eric. Consistency is good, but not if it adds confusion.

ezimanyi: +1 to Maggie's concern about confusion with the "Deployment" term in Kubernetes...though maybe "deployment strategy" is a common enough word it stands to itself?  I'd be fine with either...or we could jump right to "Traffic Management" to better encompass dark deploy.

ezimanyi (approved):
Looks great! :tada:

guides/user/kubernetes-v2/rollout-strategies/index.md
maggieneterval: Agreed, the `valid for ReplicaSets only` text links [here](https://www.spinnaker.io/guides/user/kubernetes-v2/traffic-management/#you-must-use-replica-sets) which explains "why not deployments", lmk if you think that's enough or we should highlight it more somehow?

guides/user/kubernetes-v2/rollout-strategies/index.md
maggieneterval: you guys make me laugh. updated to `red/black`.

guides/user/kubernetes-v2/rollout-strategies/index.md
maggieneterval: Hmm, if highlander isn't there, maybe it's meant to just illustrate a few common patterns rather than be an exhaustive list of all patterns implemented by all cloud providers? Not sure how that image was created but looks like the commit was made by @skim1420, so will defer to him if he would like us to update it!

guides/user/kubernetes-v2/rollout-strategies/index.md
ethanfrogers: I agree that `Deployment` is overloaded and could be confusing. I like Rollout Strategies though, I have a _slight_ concern that it could be confused with a Deployment's `strategy` thus making the Deployment/ReplicaSet thing even _more_ confusing.

I think I prefer Traffic Management _most_ but we can always fallback to that if we find that we get a lot of questions about what Rollout Strategy means.

maggieneterval: My only concern with using Traffic Management is it seems like that's only one dimension of the rollout (the other being the creation/destruction of workloads), but yeah let's definitely keep an ear to the ground and record any confusion with the terminology. Going to merge this as-is for now and happy to make any future updates going forward based on feedback!

PR #1329, opened 2019-04-25
docs(openstack): remove openstack references

seanpeters86:
Removing OpenStack provider docs as per https://github.com/spinnaker/spinnaker/issues/4316


emjburns:
LGTM, it might be good to add something linking to the RFC on the docs site. I want people to understand that we had support and we removed it because it was not maintained, but if someone wants to take over ownership we'd be happy to revert this and help them.

seanpeters86:
Good idea!

> *Note:* The OpenStack provider was supported for a period of time, but after several releases without support [this RFC](https://github.com/spinnaker/spinnaker/issues/4316) concluded with the removal of the provider from Spinnaker. If you are interested in adding this provider back in and supporting it we would be more than happy to help revert the removal.

Added the above block in, thoughts? @emjburns

emjburns:
@seanpeters86 perfect! Thank you!

emjburns (approved):

PR #1343, opened 2019-05-09
fix(customJobs): add bit about target version

ethanfrogers:
there's been a bit of confusion around when this feature would be available so I thought it best to state it explicitly.


guides/operator/custom-job-stages/index.md
```diff
@@ -7,7 +7,7 @@ sidebar:
 
 For users of Kubernetes or Titus, Custom Job Stages offer a native solution for enhancing Spinnaker's functionality with the power of containers. Custom Job Stages enable operators to extend Spinnaker without having to maintain a custom build of Orca.
 
-*Note: Custom Job stages are only supported for the Kubernetes and Titus providers.*
+*Note: Custom Job stages are only supported for the Kubernetes and Titus providers and will be available in version 1.14.*
```
dorbin, line 10: @ethanfrogers Sorry this has languished for almost a month. Given that 1.14 is released, is this obsolete?
Or maybe change it to "...and are available starting with version 1.14."?

ethanfrogers: I think this can probably be closed at this point.

PR #1361, opened 2019-05-20
Add documentation on new GCB functionality

ezimanyi:


setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
 gcloud services enable cloudbuild.googleapis.com
 ```
 
-## Configure Spinnaker to Listen for Google Cloud Build Pub/Sub Notifications
-
 Google Cloud Build sends [Build Notifications](https://cloud.google.com/cloud-build/docs/send-build-notifications)
-when the state of your build changes. You can create a pipeline trigger to invoke a pipeline based
-on the status of your build.
+when the state of your build changes.  Spinnaker subscribes to these pub/sub messages so that in can track the status
```
dorbin:
```suggestion
when the state of your build changes.  Spinnaker subscribes to these pub/sub messages so that it can...
```

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
+of builds it has initiated in a GCB stage and to trigger pipelines based on build status changes.
```
dorbin:
```suggestion
* track the status of builds it has initiated in a GCB stage
* trigger pipelines based on build-status changes
```

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
     gcloud pubsub subscriptions create $SUBSCRIPTION_NAME \
       --topic projects/$PROJECT_ID/topics/cloud-builds \
       --project $PROJECT_ID
     ```
+    
+Finally, you will need a service account that has both Cloud Build Editor and Pub/Sub Subscriber permissions, which is
+assumed to be in a file at path `$SERVICE_ACCOUNT_KEY` in the below commands.
 
-2. Configure and deploy Spinnaker via Halyard with your newly created subscription.
+## Configure Spinnaker for Google Cloud Build Support
```
dorbin:
```suggestion
## Configure Spinnaker to work with Google Cloud Build
```

setup/ci/gcb.md
```diff
@@ -8,7 +8,9 @@ sidebar:
 {% include toc %}
 
 Setting up [Google Cloud Build](https://cloud.google.com/cloud-build/) as a Continuous Integration (CI)
-system within Spinnaker enables triggering pipelines when builds complete.
+system within Spinnaker allows you to:
+ * trigger pipelines when a GCB build completes
+ * add a GCB stage to your pipeline
 
 ## Prerequisites
```
dorbin, line 15: So for prereqs, you need...
* a GCP project
* GCB API enabled (maybe we can keep these two combined in one bullet)
* subscription to cloud-build topic
* a service account that has both Cloud Build Editor and Pub/Sub Subscriber permissions

I think these should either be bullets, with content indented under them, or subheadings.  Subhead probably best.

maggieneterval (approved):
Looks perfect!  :cloud: :hammer:

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
-## Configure Spinnaker to Listen for Google Cloud Build Pub/Sub Notifications
-
 Google Cloud Build sends [Build Notifications](https://cloud.google.com/cloud-build/docs/send-build-notifications)
-when the state of your build changes. You can create a pipeline trigger to invoke a pipeline based
-on the status of your build.
+when the state of your build changes.  Spinnaker subscribes to these pub/sub messages so that in can track the status
+of builds it has initiated in a GCB stage and to trigger pipelines based on build status changes.
 
-1. Create a Subscription object for the `cloud-builds` topic in your project:
+Create a Subscription object for the `cloud-builds` topic in your project:
```
maggieneterval, line 34: Actually one random nit: should we either append a colon to all enumerated instructions that are followed by commands, or none of them?

ezimanyi: Good point.  @dorbin , any preference or convention with regards to this?

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
     PROJECT_ID=
-    SUBSCRIPTION_NAME=gCloudBuilds
+    SUBSCRIPTION_NAME=spinnaker-cloud-build
 
     gcloud pubsub subscriptions create $SUBSCRIPTION_NAME \
       --topic projects/$PROJECT_ID/topics/cloud-builds \
       --project $PROJECT_ID
     ```
+    
+Finally, you will need a service account that has both Cloud Build Editor and Pub/Sub Subscriber permissions, which is
```
dorbin:
```suggestion
Finally, you will need a service account that has both Cloud Build Editor and Pub/Sub Subscriber permissions.
```

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
+assumed to be in a file at path `$SERVICE_ACCOUNT_KEY` in the below commands.
```
dorbin:
```suggestion
The commands below look for the service account key in a path/file defined in `$SERVICE_ACCOUNT_KEY`.
```

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
-    hal config pubsub google enable
+      --json-key $SERVICE_ACCOUNT_KEY
 
+    hal config ci gcb enable
+    
     hal deploy apply
-    ```
+```
 
 ## Configure Your Pipeline Trigger
```
dorbin, line 68: While you're in here, can you change all these numbers to `1`? No reason to hardcode them.

dorbin, line 68: Above, we're on the command line. Now, suddenly, Deck. Which is fine, but it might be a good idea to put a sentence here orienting the user. 
"Configure your pipeline to be triggered by a completed GCB build:"

setup/ci/gcb.md
```diff
@@ -20,35 +22,37 @@ You can enable the API with the following `gcloud` command:
...
```
dorbin:
```suggestion
## Configure your pipeline trigger
```

setup/ci/gcb.md
dorbin:
```suggestion
## Configure Spinnaker for Google Cloud Build support
```

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
 `gcr.io/my-project-id/my-application` (replacing `my-project-id` and `my-application` with
 appropriate values). You can then [use the produced image](/reference/artifacts/in-pipelines/)
 in downstream stages.
+
+## Configure a Google Cloud Build Stage
```
dorbin:
```suggestion
## Configure a Google Cloud Build stage
```

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
...
+
+To run a Google Cloud Build as part of a Spinnaker pipeline, create a stage of type *Google Cloud Build*.  Configure the
```
dorbin:
```suggestion
To run a GCB build as part of a Spinnaker pipeline, create a stage of type *Google Cloud Build*.  Configure the
```

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
...
+stage by selecting the GCB account to use to run the build, and entering the [build configuration YAML](https://cloud.google.com/cloud-build/docs/build-config)
```
dorbin: I think all this should be broken out into steps.

setup/ci/gcb.md
dorbin: I think this should be in /guides/pipeline/ rather than in setup. 
Everything from here on down.

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
...
+In the *Produces Artifacts* section, you may supply any artifacts that you expect the build to create in order to
+make these artifacts available to downstream stages.  Google Cloud Build supports creating either GCS or Docker image
+[artifacts](https://cloud.google.com/cloud-build/docs/configuring-builds/store-images-artifacts), either of which
+will be converted to Spinnaker artifacts and injected into the pipeline on completion of the build.
+
+While your build is executing, the stage details will provide the current status of the build and a link to view
+the build logs in the Google Cloud Console:
+![](/setup/ci/gcb_status.png)
+
+## Configuration via Pub/Sub
```
dorbin: The previous section was "Configure a Google Cloud Build stage." Is this section more like, "Configure a GCB trigger for your pipeline"?

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
...
+
+Prior to version 1.14, Spinnaker did not have built-in support for Google Cloud Build, but pipelines could be
```
dorbin, line 113: Now, as of 1.14, we do support GCB, but it's not equivalent to the former support described in this paragraph, right? Assuming I'm understanding this correctly. 
That is, this paragraph sounds like we're acknowledging the old way of doing something after saying the new way. And saying you can still do it the old way.
But aren't we doing two different things here:
* trigger from a GCBuild, via pub/sub
* spawn a GCBuild from a pipeline

If this understanding is correct, I'd change this paragraph to just introduce how to configure the trigger, removing references to before/after version number.

setup/ci/gcb.md
```diff
@@ -73,3 +77,46 @@ if the build produces a Docker image, you can add an expected artifact of type *
...
+      --project $PROJECT_ID \
+      --subscription-name $SUBSCRIPTION_NAME \
+      --message-format GCB
+
+    hal config pubsub google enable
+
+    hal deploy apply
+    ```
+
+The steps to create a pipeline trigger in this case are exactly the same except that the **Subscription Name** field
```
dorbin: exactly the same as configuring the GCB stage? or same as what?

setup/ci/gcb.md
dorbin: I say include it. But if the introducing sentence trails off, such that the list item(s) complete the sentence, I like to use an elipsis...

setup/ci/gcb.md
ezimanyi: Didn't realize you could use that syntax, thanks!

setup/ci/gcb.md
ezimanyi: Updated all instructions to be preceded by colons (as none seemed to fit the case for needing an elipsis)

setup/ci/gcb.md
ezimanyi: Added subheadings for each

setup/ci/gcb.md
ezimanyi: Spawning a GCB build from a pipeline is completely new, so there's no equivalent way of doing it prior to version 1.14.

Triggering from a GCB build could be done before via the method in this section (directly configuring pubsub) but going forward the recommended way is to configure as above.  I wanted to leave this section here for users who are not yet on 1.14, or who are upgrading to 1.14 but already have the old way configured.

Maybe it would help if I made it more clear that this is an alternate way of configuring triggering, but there's no way to start a GCB build from Spinnaker prior to 1.14?

setup/ci/gcb.md
ezimanyi: I'll change this to a link directly to the above instructions so it's clear

setup/ci/gcb.md
ezimanyi: I was wondering if that might be better for the stage instructions, but it looked like that section had mostly quick summaries rather than detailed instructions.  Let's chat about where this should live.

setup/ci/gcb.md
ezimanyi: It's really about the pre-1.14 config; changed the title to reflect that

ezimanyi:
I've gone through and addressed all the comments, leaving open the two comments that might need a further round of discussion.  As discussed, I'll merge this as is so it's live when we publish the 1.14 release notes, but can make a follow-up PR for any further changes.

PR #1367, opened 2019-05-23
docs(clouddriver): fix yaml alignment

devorbitus:
The migration `jdbcUrl` attribute should be a child of the `migration` attribute and not a sibling


robzienert (approved):

robzienert:
Thanks kindly!

Issue #1033, opened 2018-10-03
Move 'Spin CLI' from 'Guides' to 'Reference'


demobox:
![image](https://user-images.githubusercontent.com/223702/46433271-6c04cb80-c71e-11e8-9cba-0b2afbf50eb3.png)

The information under 'Guides' is more persona-focused rather than component-focused. 'Reference' seems a better place for that?

/cc @jtk54

jtk54:
A lot of the content under `Spin CLI` are how-to(s), so I don't know if `Reference` is the best place for it. That being said, I don't have a problem splitting out content from the guide that can be used as reference and putting it in `Reference`. Deferring to @dorbin's best judgement there.

dorbin:
Yeah, if I can split the referency stuff from the how-to stuff without making it too much work to use the separated docs, then I'll do that. Else I'll either keep it where it is or move it wholesale, depending on how I feel after spending some time analyzing it. 

@demobox Note that I'm thinking of this based on what types of content should go in a reference, rather than the model of what types of content we have in there now. (For example, all our Halyard docs are in reference&mdash;not just the command reference.)

demobox:
> A lot of the content under Spin CLI are how-to(s)

Hm...would this split makes sense?

* Install and Configure Spin CLI to `Reference`
* Manage Applications and Manage Pipelines overviews to `Reference`
* Other "How To" sections to `User How-Tos` with titles such as "Manage Applications via the CLI"? Or include in appropriate existing articles?

lwander:
Install & configure doesn't feel like reference material to me, given that all other install & configure lives in Setup.

Manage Applications & Pipelines feels more like a guide, depending on how user-focused it is.

dorbin:
What seems referencey is the config material: `/.spin/config`. One could also make reference content from the commands themselves. But they way it's written now ("Managing your application's lifecycle," "Delete a single application with `delete`") it's more of a guide. That is, how do I do ______? rather than how does the _______ command work?

The way I feel about it right now, I think it should all remain in guides, but in time we should create a command reference.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
[...]

PR #1390, opened 2019-06-06
docs(spin): moves setup from spin u.g. to other config

dorbin:
#1033


demobox (approved):
Looks good to me, thanks!

PR #1398, opened 2019-06-12
fix(oauth): add missing 'to'

mrampton:


dorbin (approved):
Good catch, @mrampton . Thanks!

PR #1400, opened 2019-06-14
docs(custom-job-stage): add missing label parameter to example snippet

error418:
added missing `label` property to example configuration snippet


ethanfrogers (approved):

ethanfrogers:
Nice catch, thanks!

PR #1399, opened 2019-06-13
fix(docs): fix viewing {{ templateKey }} within this static site generator

milesmatthias:
The current docs result in an empty `''` here: https://www.spinnaker.io/guides/user/kubernetes-v2/run-job-manifest/#link-to-external-system because the template isn't escaped.

PR #1401, opened 2019-06-14
fix(docs): actually fix templateKey template rendering with jekyll

milesmatthias:
Fixes how `{{ templateKey }}` shows up on:

* https://www.spinnaker.io/guides/user/kubernetes-v2/annotations-ui/
* https://www.spinnaker.io/guides/user/kubernetes-v2/run-job-manifest/

Screen Shot 2019-06-14 at 9 10 12 AM

1 PR with 1 commit -- look at that! (#1399 mess)

cc @dorbin @ethanfrogers


dorbin (commented):
@milesmatthias 
I'm not an expert on the syntax, but I see differences between how youve done in in these two files. In /annotations-ui/index.md there are no spaces between the {{ and the ", but in /run-job-manifest/index.md, there are spaces. If that matters.

milesmatthias:
I tested this and it doesn't matter, it renders the same on the page, but I updated it for consistency. Thanks!

ethanfrogers (approved):
Thanks for the fix!

PR #1407, opened 2019-06-18
docs(aws) - adding the aws sig to the website

imosquera:
This add the AWS SIG to the website


robzienert (approved):

robzienert:
I _THINK_ we'll want to do some refactoring of the website here so that the community repo is the source of truth / only location for the SIGs so we don't have drift. Thoughts @spinnaker/sc @spinnaker/toc ?

aglover:
Yes, we need to point spinnaker.io to the repo. I'm ok w/some duplication until we straighten it out.

PR #1410, opened 2019-06-20
fix(jobs): document syntax for using slash in annotations

ethanfrogers:
Because jobs are defined via configuration (Java properties), the `/` character cannot be used without special syntax. Documenting this here for the future.

### DIFF
--- a/guides/operator/custom-job-stages/index.md
+++ b/guides/operator/custom-job-stages/index.md
@@ -45,6 +45,8 @@ The following properties are supported for the `kubernetes` cloud provider.
 * `application` - Name of the application to associate the job with.
 * `manifest` - YAML definition for the [Kubernetes Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) that will be run when the stage is executed.
 
+_Note: If you're using the slash (`/`) character in any manifest annotations you'll need to use this [special syntax](https://github.com/spring-projects/spring-boot/issues/13404#issuecomment-395307439) to prevent the slash from being dropped when the application reads in it's configuration. For example, if you're annotation key is `iam.amazonaws.com/role` you'll need to define it like so `[iam.amazonaws.com/role]`.
+
 ### Custom Job Stages - Titus
 
 * `cluster` - Definition of the cluster to be created when the stage is executed.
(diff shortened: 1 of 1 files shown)


ethanfrogers:
spinnaker/spinnaker#4465

maggieneterval (approved):
Very interesting, thanks for highlighting this @ethanfrogers! @dorbin is OOO until Tuesday so I'm going to go ahead and merge this just so we can point users to this documentation ASAP, but Dave feel free to suggest any wording changes after the fact and I will open a PR to amend this!

guides/operator/custom-job-stages/index.md
```diff
@@ -45,6 +45,8 @@ The following properties are supported for the `kubernetes` cloud provider.
 * `application` - Name of the application to associate the job with.
 * `manifest` - YAML definition for the [Kubernetes Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) that will be run when the stage is executed.
 
+_Note: If you're using the slash (`/`) character in any manifest annotations you'll need to use this [special syntax](https://github.com/spring-projects/spring-boot/issues/13404#issuecomment-395307439) to prevent the slash from being dropped when the application reads in it's configuration. For example, if you're annotation key is `iam.amazonaws.com/role` you'll need to define it like so `[iam.amazonaws.com/role]`.
```
maggieneterval:
```suggestion
_Note: If you're using the slash (`/`) character in any manifest annotations, you'll need to use this [special syntax](https://github.com/spring-projects/spring-boot/issues/13404#issuecomment-395307439) to prevent the slash from being dropped when the application reads its configuration. For example, if your annotation key is `iam.amazonaws.com/role` you'll need to define it like so: `[iam.amazonaws.com/role]`.
```

maggieneterval: Sorry just a couple random grammar nits before I merge this!

ethanfrogers: nice catch! thank you!

derek9gag:
@ethanfrogers @maggieneterval - Very minor one but the trailing `_` is missed at the end of the paragraph. I think it's supposed to be in italics.

ethanfrogers:
🤦‍♂ thanks for catching that @derek9gag!

PR #1404, opened 2019-06-17
chore(docs) Documentation of Amazon SQS-SNS based pubsub

clanesf:
Initial draft of Amazon SQS-SNS pubsub docs. Please give me any feedback!

PR #1421, opened 2019-06-25
chore(docs) Amazon pubsub added to nav

clanesf:
@dorbin - Added Nav as well as per your request on https://github.com/spinnaker/spinnaker.github.io/pull/1404


clanesf:
@emjburns - @dorbin asked me to add this so that it would show up in nav

dorbin (approved):
Thanks, @clanesf

PR #1422, opened 2019-06-26
Improve formatting for better syntax highlighting

nagelflorian:
Hey there, I just noticed the slightly off JSON formatting in this code block. This should improve the syntax highlighting a bit, although it's of course still invalid JSON due to the comments. I was thinking perhaps adding a basic Markdown linter and formatting tool like [Prettier](https://prettier.io/blog/2017/11/07/1.8.0.html#markdown-support) for the future would also be beneficial. That's a separate discussion though. 🙂


dorbin (approved):
@nagelflorian 
Thanks for catching these. Also, created #1423 to deal with comments in JSON.

PR #1427, opened 2019-06-28
feat(md): minimal docs

emjburns:
These are very very minimal docs meant to be hidden and used in a tooltip with a direct link.

@dorbin I'm purposefully not adding these to the general nav, and they're super rough. I'll be looking for your advice and guidance in the future once we polish these. I'm just tagging you on this so you're not surprised!


reference/managed-delivery/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Then why do I have code checked into a repo?
+
+To help you keep your resources in a known good state.
+You must publish these resource configs to Spinnaker every time they change in order for that change to be reflected in Spinnaker.
+
+### What if I want to stop managing this resource declaratively?
+
+Send us a `DELETE` for the config of this resource. 
+We will stop managing the resource, and you can make changes like normal.
```
erikmunson, line 47: maybe add that the infra itself won't be deleted? It's kind of implied but not explicitly stated.

erikmunson (approved):

PR #1431, opened 2019-07-05
docs(cf): Updated hal add account fields

pdelagrave:
Added docs about `--skip-ssl-validation` and updated the existing field that have been renamed


pdelagrave:
@dorbin

dorbin (approved):

PR #1446, opened 2019-07-10
Change suggested labels for traffic management

demobox:
- Service has a "synthetic" label selector that indicates that it doesn't actually know what it will be pointing it
- ReplicaSet has labels that more closely reflect what a "normal" ReplicaSet would look like


maggieneterval:
cc @ezimanyi

maggieneterval (approved):

ezimanyi (approved):

dorbin (approved):

PR #1462, opened 2019-07-23
fix(docs): Incorrect path in the manifest logs metadata value

pdelagrave:


pdelagrave:
The interpolation context is of `IManifest` and not directly the manifest.
The docs should whether reflect this or the deck code be corrected to make it `IManifest.manifest` instead.
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx#L54

```javascript
export interface IManifest {
  name: string;
  moniker: IMoniker;
  account: string;
  cloudProvider: string;
  location: string;
  manifest: any;
  status: { [key: string]: IManifestStatus };
  artifacts: IArtifact[];
  events: IManifestEvent[];
}
```

pdelagrave:
@ethanfrogers

ethanfrogers (approved):

PR #1470, opened 2019-07-31
chore(docs-stackdriver): Note regarding Spinnaker Monitoring - Stackdriver Custom Metrics

clanesf:
Hey guys, if this gets rejected I'll completely understand, or understand if the wording needs to be corrected, but here is my experience/thinking. 

7/30/2019 during the Kubernetes V2 SIG, I related an experience that I had when trying to use the spinnaker-monitoring custom metrics directly out of the daemon and use it to create Spinnaker Monitoring Custom Metrics in Stackdriver Monitoring. 

Basically after going and installing all of the metrics from the spinnaker-monitoring daemon in half a dozen GKE clusters running spinnaker, then using the Stackdriver API Key to create Dashboards, I had two issues

1) Almost none of the `monitoring-daemon` sidecars correctly exported the metrics so that dashboard data would show up that was actually useful. This was after a couple of weeks of figuring out how to get the API Key from our Google rep.

2) All of a sudden our Spinnaker Monitoring bill grew HUGE! Like to monitor the default metrics caused more than running all the clusters nodes in GCE (by a multiple!)

After sharing that, @maggieneterval @benjaminws @ethanfrogers and a few others related that they had similiar experiences.

My thought in adding this commit is that if we want people to adopt Spinnaker, I'd hate to see a small to medium startup get hit with those kind of unexpected expenses and just decide to abandon spinnaker (or worse) because they were using what was on the website and therefore was assumed to be equal to the Prometheus or Datadog options.

Right now, I don't know that anyone is using the direct reporting of the custom metrics to Stackdriver Monitoring (though I could be wrong, I'm open to dicussion here). I've also tried to phrase this in a way that doesn't degrade Google at all. On the contrary, I'm a bit of a Google Fanboy, I'm just more of a Spinnaker fanboy and even more of a transparency and honesty Fanboy :)  

If the language needs to be addressed in any way or we just want to go a different direction here I'm totally fine with that, just trying to save somebody some of the same time and hassles and $$ that we went thru :)


maggieneterval:
Hey @clanesf, thank you so much for the thoughtful PR. I really appreciate you following up on the conversation we all had at the Kubernetes SIG with documentation that could help prevent newer users from hitting the same unexpected expenses when setting up Stackdriver monitoring as described in our docs. This is an area of Spinnaker I know next to nothing about, so am going to tag a few other folks for thoughts: @perek, @dorbin, @ttomsu, @ezimanyi, @milesmatthias, @ewiseblatt, @duftler  -- thanks in advance for any input!

milesmatthias:
I haven't personally used SD to monitor Spinnaker, so I can't speak to the content of this PR, but I agree with its intentions -- being upfront about potential issues you may run into will save a lot of companies energy & frustration if they can get all of the possible issues up front when making architecture decisions. 

Even at large companies, this wouldn't be fun to go through and has the potential to lower trust, increase maintenance sensitivity, and create financial concerns. Big budgets can actually mean more or less price sensitivity, depending on context.

So 👍

ezimanyi:
This seems reasonable to me, and better than the potential surprises that users might encounter without this warning.  We should definitely think about how we could make this setup better for Stackdriver users, but in the meantime I think that being transparent about the issues users have encountered is a good idea.

setup/monitoring/stackdriver.md
```diff
@@ -4,9 +4,26 @@ title:  "Stackdriver"
...
   nav: setup
 ---
-
 {% include toc %}
 
+### Regarding using Spinnaker Monitoring metrics directly as Stackdriver Monitoring Custom Metrics
+
+As of July 2019, it should be noted that multiple teams running Spinnaker on clusters in GKE have have opted to export metrics to DataDog, Prometheus or other tools rather than using the spinnaker-monitoring daemon to create StackDriver Custom Metrics from the spinnaker-monitoring daemon as described below.
+
+The reason for this is that the spinnaker-monitoring service was not initially designed to run on Stackdriver Monitoring and contains a large number of metrics. Google bills for each of these Custom Metrics individually, which can add *substantial* cost with the number of metrics present in the daemon. It's also problematic that installing the default dashboards in the daemon directly into a Stackdriver Monitoring Dashboard requires a manual step of communicating with Google to get access to the Stackdriver Dashboard API ([as described below](#installing-the-stackdriver-dashboards)).
```
ttomsu, line 13: A small quibble here - it in fact _was_ designed with Stackdriver in mind (by @ewiseblatt), and he noted repeatedly (internally at least) that the cost to use it would be enormous (as you found out), and as such, added support for Prometheus and DataDog.

ttomsu (approved):

setup/monitoring/stackdriver.md
ewiseblatt: I am not going to object to the wording here.

I am responding to background and to remind some courses of action about the issue.
To travis' point, I spent a few years raising these concerns and various other issues and constraints with them. The original implementation made some undesirable cost tradeoffs in order to work around other stackdriver issues and limitations at the time just to make it possible. Many of those have since been addressed.

Earlier this year I had made a proposal with the following reference implementation to the spinnaker community as a workaround for a few general issues, which was also motivated by mitigating stackdriver issues including costs. It would also make it feasible to create builtin metrics[*] in the future, which would remove all the custom metric costs, though getting approvals for builtin metrics has its own challenges at this time (maybe less so if there were community buy in)

https://github.com/ewiseblatt/spinnaker-monitoring/tree/experimental_transform/spinnaker-monitoring-daemon/experimental

I added the mechanisms that branch relies on into the spinnaker mainline several releases ago (1.11 or 1.12). That branch is just configuration that takes advantage of the mechanisms added to support it. There are no runtime code changes required. The enabling code changes are already in the product since Q1 if not since last year.

It greatly cuts down on the number of custom metric descriptors required and provides the means to control what you want on top of that. As you point out, custom metric descriptors are where the stackdriver costs come from. If I recall I had everything I wanted with just over 100 metric descriptors (Spinnaker internals and GCE, GKE, minimal AWS in a HA deployment) as compared to well over 600). Stackdriver gives you the first 500 free so depending on how much you are using elsewhere, that wouldnt add anything.

BTW, it might be the case that if you use prometheus with stackdriver then they do not charge for custom metrics. This is because the prometheus metrics are put into a different namespace so are technically not custom metrics (which impacts engineering tradeoffs because custom metrics are a more costly resource).

Having said that, if you are free to pick anything to monitor with, I like straight prometheus from among those choices. But if you do need to use stackdriver (e.g. you are monitoring other things with stackdriver and want a single monitoring system) then I think the proposal is the way to go. Personally would use the proposal regardless of metric store if you arent tied to the existing data model.

The branch could be moved into the mainline. The reasons it wasnt are not technical but this isnt the place to go into it. However if there was community buy in then I think it could be feasible to make the stackdriver costs go to 0 adding builtin metrics but there are a lot of policy hoops to jump through to get there.

[*] "builtin metrics" is not the current stackdriver term. I forget what they call them. By "builtin metrics" I mean the kind that are standard metric descriptor across all projects, not "custom metrics" which are specific metric descriptor to a particular project.

clanesf:
@ttomsu @ewiseblatt - Thank you guys for taking the time to read this and providing the background. What I wrote was an assumption based on my limited experience, so I'm happy to know the true story. 

I'm happy to amend this pull request with some updated information regarding the history here and include a link to the repo Eric linked above if you think that's a better way of handling this. I just want to provide the best experience possible for new adopters of the project.

ewiseblatt:
I wouldnt link from spinnaker.io into my personal project. I can talk to
plumpy tomorrow.

setup/monitoring/stackdriver.md
iniinikoski: Hi,

I noticed this discussion and need to comment as well... We are not using Stackdriver but Datadog - and the amount of custom metrics Spinnaker is sending is just enormous. (almost 30000 custom metrics a month), creating an extra invoice of $1000- about. Just wanted to say that Stackdriver's not the only one to "worry"...

I remember @ewiseblatt discussing this with you in Slack - and you pointed me to your experimental. I gave it a very quick go - but it broke basically all Spinnaker metrics in Datadog dashboards :) I did not have time to try to sort those out.

Going back to the roots of the issue - isn't there anything to be done - at "it's source", so the amount of custom metrics wouldn't be so explosively high - by default? It really is not a great feedback if someone just plugs in the monitoring - and you receive an interesting invoice next...

ewiseblatt:
The experimental renames all the metrics which is why the dashboards break.
The original strategy was working around another stackdriver limitation on
the number of tags you can have on a metric so we made each service have
its own metrics. That was also before HA which blew everything up. The
experimental change addresses that by consolidating metrics and reshaping
their tags and giving control over what's stored.

I had an earlier attempt for "metric filtering" as the original means for
mitigating which metrics are captured in case we hit the situation you are
reporting. That is still in there but deprecated. It is based on regular
expressions of metric names you want to keep or want to discard and might
be able to throw away some tags. I dont recall. I never documented it
because was never satisfied with it to have to support and wanted to end up
somewhere closer to the proposal I made.

I would suggest looking at the experiment again and building datadog
dashboards for it. I created prometheus dashboards along with it so should
not be that bad. The original datadog support (I think 3 years ago) was
only a few hours to put together -- their API was really easy and straight
forward.

I would also use the metric explorer in the metric server to see what
metrics are available. If you hit the home page of the server there will be
a menu with a link taking you there. If you have a transformation
configuration such as the one I give as an example then the explorer will
show you which metrics are being discarded as well as how they where
transformed. You can use that to refine which you care about and which not.

Another thing you could try is the following options in the daemon
spectator:
  # This also implies undecorated metric names so is not backward compatible
  inject_service_tag: true
  decorate_metric_name: false

This too is going to break the existing dashboards (because names are
different) and is less dramatic in affecting cost. But the metric names and
tags wont be as dramatically different since it doesnt change how metrics
are organized or try to standardize terminology or patterns.
What this will do is add the service name (and HA variation) as additional
tags on the metric rather than making them distinct names.
It wont affect your volume of metrics, but will change the number of
distinct metric names you have. I dont know how datadog charges, but that
might offer some relief.
For stackdriver it might throw you over its tag limit for some metrics. The
transformations in experimental reshape the tags to avoid that.

317brian:
There was a good amount of discussion on this, but it seems to have died down. @ewiseblatt and @iniinikoski, are you okay  with me merging this since @ttomsu signed off?

ewiseblatt:
Sorry, I assumed this had already been merged. I'm taking a last look at
the overall PR and it seems that it isnt synced to head. If not can you
update it one last time?

iniinikoski:
> There was a good amount of discussion on this, but it seems to have died down. @ewiseblatt and @iniinikoski, are you okay with me merging this since @ttomsu signed off?


👍  Sure, anyway my feedback as more in higher level I think, not Stackdriver-specific...

PR #1474, opened 2019-08-06
fix(minio): Make note about running on localhost

jervi:
If using `http://localhost:9001`, the AWS SDK will try to connect to `http://<bucket-name>.localhost:9001`, which will not work.

### DIFF
--- a/setup/install/storage/minio.md
+++ b/setup/install/storage/minio.md
@@ -22,7 +22,7 @@ on an endpoint reachable by Spinnaker. Record the following values:
 
 * `ENDPOINT`: The fully-qualifed endpoint Minio is reachable on. If Minio is
   running on the same machine as Spinnaker, this might be
-  `http://localhost:9001`.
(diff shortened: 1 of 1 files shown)


setup/install/storage/minio.md
```diff
@@ -22,7 +22,7 @@ on an endpoint reachable by Spinnaker. Record the following values:
 
 * `ENDPOINT`: The fully-qualifed endpoint Minio is reachable on. If Minio is
   running on the same machine as Spinnaker, this might be
-  `http://localhost:9001`.
+  `http://127.0.0.1:9001` (note that using `localhost` instead of `127.0.0.1` generally won't work).
```
dorbin: s/generally won't/doesn't/

In fact, the explanation you added as  a comment in the PR might be good to put in the doc.

jervi: Will fix, thanks!

dorbin (approved):
Thanks for the update, @jervi !

PR #1476, opened 2019-08-07
Fix double quotes

ryu1kn:
Fixed double-quote characters so that we can start trying pipeline template by copy & paste the sample json.

Also removed space character for consistency and added missing comma.


louisjimenez (approved):

dorbin (approved):
Thanks @ryu1kn

PR #1475, opened 2019-08-06
Update artifacts Reference for new artifacts UI

fifthposition:
This adds an updated version of the artifact Reference documentation, showing and describing the new UI enabled with the `artifactsRewrite` feature flag.

The existing artifact Reference is left alone. The new artifact Reference is not in the navigation. To see the new version of an artifact Reference page when viewing the site, you can add `-with-artifactsrewrite` to the `/artifacts` in the URL. For example, for the path `/reference/artifacts/types/docker-image/`, you can substitute `/reference/artifacts-with-artifactsrewrite/types/docker-image/`.

This also adds one new page ("In Cloud Foundry"), which is at `/reference/artifacts-with-artifactsrewrite/in-cloud-foundry/`.

cc @dorbin @claymccoy @okundzich

PR #1482, opened 2019-08-09
'artifactsRewrite' artifact Reference: additions

fifthposition:
These are just things I missed in my last PR, #1475.

• Add 'Maven Artifact' to “New UI” navigation list of artifact types
• Add missing info on use of artifacts in pipeline stages (you can use either a previously-defined artifact, or define a new artifact inline--our screenshots all show inline, I think, but we need to mention the other possibility)

cc @dorbin


dorbin (approved):
LGTM
Thanks Ben

PR #1491, opened 2019-08-28
Add documentation for dynamic configuration (Config Server)

fifthposition:
This adds documentation (under a new "Reference / Configuration" section) for dynamic configuration, which uses the Spring Cloud Config project.

cc @dorbin @scottfrederick


dorbin (commented):
Ben, what do you think about moving this to /setup/ rather than reference? There's certainly a lot of referencey material in here, but also a lot of how-to, and because we already have configuration in /setup/, I think this might be good in there. And we can think about how to divide up the reference material from the how-to material, for this and other docs, as part of our doc reorg.

PR #1492, opened 2019-08-30
fix duplicated words for traffic-management readme

chy168:


dorbin (approved):
Thanks for the fix, @chy168 !

PR #1500, opened 2019-09-12
feat(docs): Note for path-style-access with Minio

pdelagrave:
Common gotcha among people installing Spinnaker

### DIFF
--- a/setup/install/storage/minio.md
+++ b/setup/install/storage/minio.md
@@ -68,6 +68,9 @@ echo $MINIO_SECRET_KEY | hal config storage s3 edit --endpoint $ENDPOINT \
 hal config storage edit --type s3
 ```
 
+## Notes
+> :warning: Unless Minio is configured to [support virtual-host-style requests](https://docs.min.io/docs/minio-server-configuration-guide.html#Domain), the path style access mode will have to be enabled: `hal config storage s3 edit --path-style-access true`. Front50 will otherwise refuse to start and throw `java.net.UnknownHostException: $BUCKET_NAME.localhost` exceptions.
+
 ## Next steps
 
 After you've set up Minio as your external storage service, you're ready to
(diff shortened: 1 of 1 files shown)


setup/install/storage/minio.md
```diff
@@ -68,6 +68,9 @@ echo $MINIO_SECRET_KEY | hal config storage s3 edit --endpoint $ENDPOINT \
 hal config storage edit --type s3
 ```
 
+## Notes
+> :warning: Unless Minio is configured to [support virtual-host-style requests](https://docs.min.io/docs/minio-server-configuration-guide.html#Domain), the path style access mode will have to be enabled: `hal config storage s3 edit --path-style-access true`. Front50 will otherwise refuse to start and throw `java.net.UnknownHostException: $BUCKET_NAME.localhost` exceptions.
```
dorbin:
```suggestion
> :warning: Unless you've configured Minio to [support virtual-host-style requests](https://docs.min.io/docs/minio-server-configuration-guide.html#Domain), you need to enable the path-style access mode: `hal config storage s3 edit --path-style-access true`. Otherwise, Front50 will fail with `java.net.UnknownHostException: $BUCKET_NAME.localhost` exceptions.
```

dorbin (commented):
@pdelagrave 

Some small wording suggestions. 

What do you think about either...
* adding something to the above section on installing Minio saying to configure virtual-host-type requests
or
* Adding a step somewhere saying to enable path-style access mode using `hal`?

Is one preferred over the other.

pdelagrave:
@dorbin 
I don't have a strong opinion on where it should be displayed in the page structure as long as the user understand it's an optional step depending on how Minio was configured by the user. Feel free to move it where you think it belongs.

dorbin (approved):

PR #1502, opened 2019-09-12
Use Updated AWS Node Group File

drewpc:
Updated the URL to use the official Amazon (updated) node group file.


drewpc:
Disregard.  More changes required.

PR #1506, opened 2019-09-16
deprecate 1.14.14

louisjimenez:


plumpy (approved):
I just merge these without a review, honestly :grinning: 

We should maybe add to the doc that doing so is allowed...

louisjimenez:
Great, I updated the doc to reflect that.

PR #1507, opened 2019-09-17
Fix entityTags guide for Spinnaker 1.16

vava:
Spinnaker 1.16 expects entityTags object in settings, otherwise it can't load an applications UI.


plumpy (approved):
Thanks for fixing this. It seems like a bug that this is required in your given, but given entity tags' relative lack of use and our lack of knowledge about them, I don't think it's something we'll have time to fix.

PR #1524, opened 2019-10-02
docs(plugins): plugin users docs

bpowell:
Documentation for how plugin users can install and configure plugins


317brian:
I made some changes and added xlinks between the user and creator docs since they're in separate sections.

PR #1538, opened 2019-10-16
fix(dev/kork): update command for publishing to maven local

ethanfrogers:
The previous version of this command didn't work out of the box. I asked around and found that you have to manually override the property to enable publishing.


ezimanyi (approved):
Thanks for fixing this!

PR #1542, opened 2019-10-17
feat(docs): Add JPMC and GroundTruth logos

stuposluns:


317brian (approved):
Built it locally, and it LGTM. None of the logos are broken.

PR #1558, opened 2019-10-28
docs(spin/canary): Adds a few fixes for canary cc cmds.

jtk54:
Fixes up broken formatting. Note that this page is also not discoverable yet. We should probably hide it until the canary commands make it into a proper `spin` release.


jtk54:
@louisjimenez

louisjimenez (approved):

PR #1566, opened 2019-10-30
docs(plugins): clarify what the enabled parameter expects

317brian:


317brian:
@link108 can you take a look when you get back from vacation or @bpowell if you get a chance before then

317brian:
This is autogenerated, so the source files need to be updated.

PR #1567, opened 2019-10-30
feat(docs): add 4 new logos

stuposluns:
Added Avast, SAP, TransUnion and Comcast


317brian (approved):
Built locally and nothing looks broken.

PR #1569, opened 2019-10-31
chore(artifacts/gitRepo): git repo artifact

ethanfrogers:


ethanfrogers:
@brian-armory @dorbin a couple of questions:
1. This feature is only available in 1.17 and I'd like to put a note at the top indicating that. Do we have a style we like for things like that?
2. For the Artifacts Rewrite version of this doc the table of contents makes the top section look a bit off. It pushes the table down below it. Do you have any suggestions for this? It looks like it only works for other pages because there are fewer sections.

image

317brian:
For 1) I haven't noticed a standard style. (A style guide is one of the things the Docs sig wants to come up with.)

I recommend inserting

```
<div class="notice--info">
  <strong>Note:</strong> This feature is only available in version 1.17 or later.
</div>
```

That'll put it in a box.

You can also use the {: .notice--info} syntax instead.

https://mmistakes.github.io/minimal-mistakes/docs/utility-classes/#notices for more info. 

--
I made some minor copyedits. I'll look into the theme some more to figure out if we can do anything about the table/TOC issue unless @dorbin or @fifthposition know 
 off the top of their heads?

fifthposition:
For 1), I had a similar concern for our external configuration page in "Setup". We used a simple notice on that page (we had to explain that the feature came to different Spinnaker services in different versions):

```
> For Clouddriver, support for external configuration was added in Spinnaker 1.15. For Echo and Igor, support was added in Spinnaker 1.16.
```

For 2), yes @ethanfrogers I think the TOC only works on other pages because they have fewer sections. I’m not sure of a way around that. I probably would just put up with it (or do something drastic like restructure the content of the page to have more text up front or use fewer subheadings) for now, unless @brian-armory has found anything else we can do using the theme to get around it.

reference/artifacts-with-artifactsrewrite/types/git-repo/index.md
```diff
@@ -0,0 +1,101 @@
...
+Git Repo artifacts are references to Git repositories that are hosted by a Git hosting service. They are consumed
+by stages that need multiple files to produce an output such as the Bake (Manifest) stage when using the Kustomize template
+renderer. Unlike other artifact implementation, the Git Repo artifact will work with any Git hosting service as long as the
+repository can be cloned using the Git CLI.
+
+## Git Repo artifact in the UI
+
+The pipeline UI exposes the following fields for the GitHub file artifact:
+
+<table>
```
dorbin: Could we do this using a Markdown table, like the other table below?

reference/artifacts/types/git-repo/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+Git Repo artifacts are references to Git repositories that are hosted by a Git hosting service. They are consumed
+by stages that need multiple files to produce an output such as the Bake (Manifest) stage when using the Kustomize template
+renderer. Unlike other artifact implementation, the Git Repo artifact will work with any Git hosting service as long as the
+repository can be cloned using the Git CLI.
+
+## Fields
+
+
+| Field              | Explaination                                                                                                                                                                                            |
```
dorbin: s/Explaination/Explanation/

dorbin (commented):
Regarding 1)
Because we don't yet have versioned docs, I think it's a good idea to put in this notice. 
I think we should prefer the Markdown syntax over html, as Brian indicated above. 

Regarding 2) I've seen this before, and I think I solved it, but I don't remember where or how. :/
I'll see what I can come up with .

dorbin:
Actually, I think where I saw it before was with an image. In this case you've got a table. See what happens if you refactor that table as Markdown?

ethanfrogers:
Thanks for the input everyone, that was really helpful. @dorbin, I converted the table to Markdown but it didn't help the TOC issue, which, makes sense, since MD is converted to HTML. I ended up changing the headings that were causing the problem to **bold** instead of actual headings. It isn't exactly consistent with other pages but seemed worth the trade off to get formatting right.

reference/artifacts-with-artifactsrewrite/types/git-repo/index.md
```diff
@@ -0,0 +1,84 @@
...
+{%
+    include
+    figure
+    image_path="./bake-manifest-stage-git-repo.png"
+    caption="Configure a Bake (Manifest) stage to use a Git Repo artifact."
+%}
+
+## Git Repo artifact in a pipeline definition
+
+| Field              | Explaination                                                                                                                                                                                            |
```
dorbin:
```suggestion
| Field              | Explanation                                                                                                                                                                                            |
```

reference/artifacts-with-artifactsrewrite/types/git-repo/index.md
```diff
@@ -0,0 +1,84 @@
...
+repository can be cloned using the Git CLI.
+
+## Git Repo artifact in the UI
+
+The pipeline UI exposes the following fields for the GitHub file artifact:
+
+| Field | Explanation |
+| ------|--------------|
+|Account| A Git Repo artifact account. |
+|URL| HTTP or SSH URL of your Git repository. Artifact accounts configured to use a username and password should use HTTP while accounts configured to use an SSH private key should use the SSH URL format. |
```
dorbin:
```suggestion
|URL| HTTP or SSH URL of your Git repository. Use HTTP for artifact accounts configured to use a username and password, and use the SSH URL format for accounts configured to use the SSH private key. |
```

reference/artifacts-with-artifactsrewrite/types/git-repo/index.md
```diff
@@ -0,0 +1,84 @@
...
+    image_path="./bake-manifest-stage-git-repo.png"
+    caption="Configure a Bake (Manifest) stage to use a Git Repo artifact."
+%}
+
+## Git Repo artifact in a pipeline definition
+
+| Field              | Explaination                                                                                                                                                                                            |
+|--------------------|--------|
+| `type`             | Always `git/repo`.                                                                                                                                                                                       |
+| `reference`        |  HTTPS or SSH URL of your Git repository. Artifact accounts configured to use a username and password should use HTTPS while accounts configured to use an SSH private key should use the SSH URL format. |
```
dorbin:
```suggestion
| `reference`        |  HTTPS or SSH URL of your Git repository. Use HTTPS for artifact accounts configured to use a username and password, and use the SSH URL format for accounts configured to use an SSH private key. |
```

reference/artifacts/types/git-repo/index.md
```diff
@@ -0,0 +1,63 @@
...
+|--------------------|--------|
+| `type`             | Always `git/repo`.                                                                                                                                                                                       |
+| `reference`        |  HTTP or SSH URL of your Git repository. Artifact accounts configured to use a username and password should use HTTP while accounts configured to use an SSH private key should use the SSH URL format. |
```
dorbin: (Same as above, in those other two instances of this)

dorbin (approved):

317brian (approved):

PR #1573, opened 2019-11-02
fix(docker): force an explicit bundler version in Dockerfile

erikmunson:
It appears that about 24 days ago a change was merged which [updated the `BUNDLED WITH` block in the lockfile to 2.0.2 up from 1.17.1](https://github.com/spinnaker/spinnaker.github.io/pull/1526/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R124-R125). Since then, trying to run the built Dockerfile in this repo has failed with an error about the lockfile requiring bundler > 2.

This seemed weird to me because the `gem install bundler` call in the Dockerfile was theoretically pulling in a newer bundler version, and the BUNDLED WITH thing in the lockfile should've made bundler use the right version. Turns out the ruby base image we're using is [setting `BUNDLER_VERSION` to some value they pick](https://github.com/docker-library/ruby/issues/246), and that means when we run `bundle exec` we're _actually_ running an old (< 2.0) version that was specified by the base image. It's always been that way, but only became a visible issue because there was a major version bump in the lockfile.

The ruby base image [just recently fixed this](https://github.com/ruby/ruby-docker-images/pull/5), but trying to upgrade feels like it's probably a rabbit hole and I'd prefer to stabilize the Dockerfile now and think about upgrading separately. This fix does two things:

- Hard code the `gem install bundler` call to get version `2.0.2` (the version in the lockfile)
- Explicitly set `BUNDLER_VERSION` to `2.0.2`

This means that in the future, we won't implicitly/invisibly be using a version we don't expect, whether via `gem install` or `BUNDLER_VERSION`. It should also be more clear to other developers what's wrong if the lockfile + Dockerfile versions drift in the future and they see a similar error about bundler versions, because when you go to the Dockerfile it's very obvious which version is in use.

**Note:** I am not good /experienced at Docker stuff in general. Please don't hesitate to suggest a totally different way of fixing, I have no idea what I'm doing and I just want to run this thing successfully 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@
 # You must mount $(pwd) to /code (this allows pages to be dynamically regenerated without having to rebuild the Dockerfile)
 FROM ruby:2.4.1
 RUN apt-get install -y git bzip2 libssl-dev libreadline-dev zlib1g-dev
-RUN gem install bundler
+RUN gem install bundler -v "2.0.2"
+ENV BUNDLER_VERSION="2.0.2"
 
 # bundle install from a different directory to the /code mount for a large performance increase (at least on Docker-for-Mac)
 COPY Gemfile* /install/


cfieber (approved):

317brian:
@erikmunson thanks for the fix! I don't know how I missed the Bundler version when I merged that PR. Mea culpa. Hopefully your work will make it more resilient to human error.

erikmunson:
no need for regret Brian! This Dockerfile has been silently doing the wrong thing (but with the right output/behavior), and we wouldn't have known without something triggering this error/failure. now we know and it's fixed!

PR #1585, opened 2019-11-07
docs(managed): add overview and reference for resource statuses

erikmunson:
This adds a page whose whole purpose is to give a deeper level of detail/clarity on resource statuses than we have room to add in the UI. There's an overview section that covers what the purpose of a status is and some guidance on how to find a resource's status, but the primary content is a detailed reference guide to help folks get to know the different statuses and provide context on what to expect + how to investigate and respond to unexpected situations.

Right now I'm providing some pointers to the task view and high level guidance like 'investigate the current state of the resource manually and double check that nothing is broken'. This is somewhat weak guidance because we don't have excellent debugging/observability tools yet, but once the history modal is built out I intend to add a similar section on using that view to debug issues with cross-linking from this page. I'll also be PR-ing the UI after this is merged to deep link to the current status of a resource on this page so folks can deep dive into the feedback they're getting in Deck.

If you take the time to look through, **please** do a pass for typos or broken/weird sentences — there's a ton of words here and I'm sure some janky stuff made it in.

(cc @gal-yardeni @luispollo since this repo doesn't let me add you as reviewers)


reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+  figure
+  image_path="./resource-status-unknown.png"
+  caption="Example of a resource with the Unknown status"
+%}
+
+When a resource has an Unknown status, it means Spinnaker was unable to determine the current status for some reason. Because the status is unknown, it's tough to tell by the status alone whether something is *wrong*, or just unexpected.
+
+After encountering this status, Spinnaker will continue trying to check the resource for its status and may eventually succeed. If it does, the resource will transition to whatever status is applicable at that time.
+
+If you see this status, it's probably best to investigate the current state of the resource manually and double check that nothing is broken. If you decide manual intervention is needed to remedy any issues you might discover, you can temporarily pause management on the Config view for your app.
```
gal-yardeni: Should we add here a screenshot of how to pause management?

gal-yardeni (commented):
Nice! super informative :)

reference/managed-delivery/resource-status/index.md
erikmunson: yeah great point — I should've mentioned in the description that I'd like to add a totally separate doc page around pausing / resuming management with screenshots, what to expect when you unpause, etc.

I opted to avoid going into detail about pausing here in favor of linking out to that doc once it's ready in a separate PR.

reference/managed-delivery/getting-started.md
```diff
@@ -7,15 +7,15 @@ sidebar:
 
 {% include toc %}
 
-You'll be interacting with Spinnaker's Managed Delivery using curl and the Spinnaker api for now.
+You'll be interacting with Spinnaker's Managed Delivery using `curl` and the Spinnaker API for now.
```
luispollo: Since you mention the UI below, should we expand this a bit to say something like “UI support for Managed Delivery is limited at this point, so for certain use cases you’ll be interacting with Spinnaker using curl...”?

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
+---
+layout: single
+title:  "Resource Status"
+sidebar:
+  nav: reference
+---
+
+{% include toc %}
+
+*If you're not sure what a Managed Resource is, check out our [overview of Managed Delivery](/reference/managed-delivery) first.*
```
luispollo: (Cosmetic) How do you feel about using :bulb: for tips like these?

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Overview
+Managed Resources will always have a **status** which describes the current state of the resource from Spinnaker's perspective. A resource's status can help you quickly answer questions like:
+  - Is Spinnaker taking automatic action on this resource right now?
+  - Does the actual configuration of this resource match the declarative configuration I gave to Spinnaker?
```
luispollo: I prefer “actual state” to “actual configuration” — I think it’s more accurate as well.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+  figure
+  image_path="./resource-status-happy.png"
+  caption="Example of a resource with the Happy status"
+%}
+
+When a resource has a Happy status, all is well and the actual state of the resource matches its declarative configuration.
+
+If Spinnaker detects a drift from the declarative configuration in the future, the resource will change status to [Diff](/reference/managed-delivery/resource-status/#diff). Once Spinnaker starts taking automatic action to correct the drift, the status will change to [Actuating](/reference/managed-delivery/resource-status/#actuating).
+
+While Spinnaker is managing a resource, manual changes made via the UI or the API will be automatically reversed ("stomped") in favor of the declarative configuration.
```
luispollo: This same paragraph appears repeatedly in most resource status descriptions. I’d suggest moving it up before the first status header as a note, and adding the word “actively” in “While Spinnaker is actively managing a resource” to address the cases when this does not apply.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Created
+{%
+  include
+  figure
+  image_path="./resource-status-created.png"
+    caption="Example of a resource with the Created status"
+%}
+
+When a resource has a Created status, it means Spinnaker was just told to start management and hasn't taken any action yet.
```
luispollo: Instead of “Spinnaker was just told to start management”, perhaps “Spinnaker has just received the resource’s declarative configuration” (or something along those lines)? I always immediately think “who told it to do anything?” when that verb is used.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+  - A problem retrieving or processing information about the current state of the resource
+  - A problem with the declarative configuration (invalid settings, etc.)
+  - A bug or issue with Spinnaker itself
+
+After an error, Spinnaker will continue trying to check the resource and may eventually succeed. If it does, the resource will transition to whatever status is applicable at that time.
+
+If you see this status, it's probably best to investigate the current state of the resource manually and double check that nothing is broken. If you decide manual intervention is needed to remedy any issues you might discover, you can temporarily pause management on the Config view for your app.
+
+
+### Unhappy
```
luispollo: Should we consider changing the order of the statuses on this page so that there’s a more logical flow for the reader? I think “Created” could be first, since it’s the first thing that happens. Then “Diff”, “Actuating”, “Happy”, “Unhappy”, “Error” and “Unknown”, in that order.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+When a resource has an Error status, it means something went wrong while trying to check the resource for drift from its declarative configuration. Because something went wrong, Spinnaker can't take automatic actions and drift from the declarative configuration won't be resolved. Some examples of problems that might lead to an error status:
+
+  - A problem retrieving or processing information about the current state of the resource
+  - A problem with the declarative configuration (invalid settings, etc.)
+  - A bug or issue with Spinnaker itself
+
+After an error, Spinnaker will continue trying to check the resource and may eventually succeed. If it does, the resource will transition to whatever status is applicable at that time.
+
+If you see this status, it's probably best to investigate the current state of the resource manually and double check that nothing is broken. If you decide manual intervention is needed to remedy any issues you might discover, you can temporarily pause management on the Config view for your app.
```
luispollo, line 157: How does one go about investigating current state and double checking nothing is broken? Should we include some troubleshooting guidance or is that out of scope for this PR?

luispollo (commented):
This is great, @erikmunson! Well written, friendly and informative! I left some comments/questions/suggestions inline. Hope they help!

reference/managed-delivery/getting-started.md
erikmunson: yep good call thanks, I'm going to mention both the UI and 'version control' since technically these resource CRUD operations will probably always be done via the command line or version control integration.

reference/managed-delivery/resource-status/index.md
erikmunson: You know I have trouble saying no to emojis

reference/managed-delivery/resource-status/index.md
erikmunson: very yes, thanks for that catch and will change. I've been trying to be disciplined about that term and as I expected writing this many words at once got me tripped up 😄

fwiw I have been making an attempt to stick to three distinct terms for the different states/configurations being talked about:
- **Declarative configuration**: the YAML or JSON you write and submit to us, might be abbreviated or delegate some decisions to defaults
- **Desired State**: the _fully realized_ settings for the thing you submitted a declarative configuration for —  all defaults filled in etc.
- **Actual state**: the settings in the real world

one of my todos that i don't have time for right now is to lay out the concept of continuous management and define those terms in a 'concepts' doc that sits parallel to these — you'll notice right now I'm making the executive decision to use 'declarative configuration' where it'd technically be more accurate to say 'desired state' because IMO we need to have a nice way to link out to a definition for desired state before we throw the term around (it feels a bit tough to parse if you're new/confused).

reference/managed-delivery/resource-status/index.md
erikmunson: Here's my dilemma:

1. I see a status in Spinnaker via the UI, I'm new to this and a bit confused about the specifics
2. I click on the 'Learn more about this' link because I'm thinking 'I don't wanna mess with anything until I know what's up with this status, what if I break it'
3. I get deep-linked right to the current status I'm seeing and I read the description for it, maybe another couple that are cross-linked from the description too
4. Now I'm feeling a bit more confident. Seems like I understand what's going on and I'm ready to be a team player with the Spinnaker machine
5. I go back to the UI and make a manual change that seems reasonable to me **because I never scrolled up and read the beginning of the reference section**.
6. My manual change looks like it's working, but then it doesn't.  Now I'm confused and it seems like this thing is broken.

Am I being too cautious? I ask that question genuinely because I'm not sure and I want to do the right thing. My greatest fear is that the feedback and context from this UI -> docs workflow _fails to prevent an avoidable incident_.

reference/managed-delivery/resource-status/index.md
erikmunson: love it, very agree.

reference/managed-delivery/resource-status/index.md
erikmunson: yes, we should IMO. I've been struggling a bit with ordering on these (I have them alphabetized in code because... I guess it felt nice?) and I think that's a better flow that what's here right now.

reference/managed-delivery/resource-status/index.md
erikmunson: Yeah this is was I was eluding to with this somewhat defeatist comment in the PR description:
> Right now I'm providing some pointers to the task view and high level guidance like 'investigate the current state of the resource manually and double check that nothing is broken'. This is somewhat weak guidance because we don't have excellent debugging/observability tools yet, but once the history modal is built out I intend to add a similar section on using that view to debug issues with cross-linking from this page. 

IMO this is ok in the short term while folks at Netflix are the only people using it and know how to reach out to us, but soon we'll need to do better.

reference/managed-delivery/resource-status/index.md
emjburns: Big fan of having consistent icons to help break up the monotony of text.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+Example on a cluster:
+{%
+  include
+  figure
+  image_path="./cluster-resource-status-ui.png"
+%}
+
+
+### API
+If you're [setup to use the API](/reference/managed-delivery/getting-started/#env-setup) and have the [ID of a resource](/reference/managed-delivery/getting-started/#find-a-resources-id) you can use it to call the `/managed/resources/<resourceId>/status` endpoint.
```
emjburns:
```suggestion
If you're [setup to use the API](/reference/managed-delivery/getting-started/#env-setup) and have the [ID of a resource](/reference/managed-delivery/getting-started/#find-a-resources-id) you can call the `/managed/resources/<resourceId>/status` endpoint.
```

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+{%
+  include
+  figure
+  image_path="./resource-status-happy.png"
+  caption="Example of a resource with the Happy status"
+%}
+
+When a resource has a Happy status, all is well and the actual state of the resource matches its declarative configuration.
+
+If Spinnaker detects a drift from the declarative configuration in the future, the resource will change status to [Diff](/reference/managed-delivery/resource-status/#diff). Once Spinnaker starts taking automatic action to correct the drift, the status will change to [Actuating](/reference/managed-delivery/resource-status/#actuating).
```
emjburns:
```suggestion
If Spinnaker detects a drift from the declarative configuration in the future, the resource status will change to [Diff](/reference/managed-delivery/resource-status/#diff). Once Spinnaker starts taking automatic action to correct the drift, the status will change to [Actuating](/reference/managed-delivery/resource-status/#actuating).
```

reference/managed-delivery/resource-status/index.md
emjburns: I agree with you @erikmunson. Let's repeat things in every section if it's meant to be a reference (exact same wording every time == "oh I've read this before I understand and I'm comfortable with the meaning"). These are new and scary concepts and will take people at least 3x of reading something to actually learn it.

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+{%
+  include
+  figure
+  image_path="./resource-status-created.png"
+    caption="Example of a resource with the Created status"
+%}
+
+When a resource has a Created status, it means Spinnaker was just told to start management and hasn't taken any action yet.
+
+If the actual state of the resource matches its declarative configuration, the resource will transition to the [Happy](/reference/managed-delivery/resource-status/#happy) status the next time Spinnaker checks it.
```
emjburns:
```suggestion
If the actual state of the resource matches the declarative configuration, the resource will transition to the [Happy](/reference/managed-delivery/resource-status/#happy) status the next time Spinnaker checks it.
```

emjburns: I freaking love the suggestion mode. My status is `HAPPY`

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Actuating
+{%
+  include
+  figure
+  image_path="./resource-status-actuating.png"
+  caption="Example of a resource with the Actuating status"
+%}
+
+When a resource has an Actuating status, it means Spinnaker detected a drift from the declarative configuration and automatic action is in progress to resolve it. You can go to the Tasks view in your app to see the actions Spinnaker is taking on your behalf, and whether they succeeded or failed.
```
emjburns:
```suggestion
When a resource has an Actuating status, it means Spinnaker detected a drift from the declarative configuration and an automatic action is in progress to resolve it. You can go to the Tasks view in your app to see the actions Spinnaker is taking on your behalf, and whether they succeeded or failed.
```

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+  include
+  figure
+  image_path="./resource-status-paused.png"
+  caption="Example of a resource with the Paused status"
+%}
+
+When a resource has a Paused status, it means Spinnaker is configured to continuously manage the resource but you've chosen to temporarily pause management. While management is paused Spinnaker won't check for drift from the declarative configuration, and it won't take any automatic actions. You can go to the Config view in your app to pause and resume management across the entire application.
+
+When you resume management, Spinnaker will check for drift from the declarative configuration. If a drift is detected, the status will change to [Diff](/reference/managed-delivery/resource-status/#diff). If no drift is detected, the status will change to [Happy](/reference/managed-delivery/resource-status/#unhappy).
+
```
emjburns, line 128: Is it worth adding a line here about how, in the future, we want to show you want resources we will act on? and, in the even more future, what actions we will take? I don't want someone to read this and think "wow that's a shitty experience" - I just want to set expectations where things will grow and change and get better.

It might be too distracting to put that here though.

reference/managed-delivery/resource-status/index.md
emjburns: Great. My previous comment about "should we talk about future un-pause plans" is not applicable for this doc then.

reference/managed-delivery/resource-status/index.md
emjburns: Never mind, i see you plan to write more words later. 👍

reference/managed-delivery/resource-status/index.md
```diff
@@ -0,0 +1,160 @@
...
+### Error
+{%
+  include
+  figure
+  image_path="./resource-status-error.png"
+  caption="Example of a resource with the Error status"
+%}
+
+When a resource has an Error status, it means something went wrong while trying to check the resource for drift from its declarative configuration. Because something went wrong, Spinnaker can't take automatic actions and drift from the declarative configuration won't be resolved. Some examples of problems that might lead to an error status:
```
emjburns:
```suggestion
When a resource has an Error status, it means something went wrong while trying to check the resource for drift from its declarative configuration. Because something went wrong, Spinnaker can't take automatic actions and drift from the declarative configuration won't be resolved (until something changes). Some examples of problems that might lead to an error status:
```
The current phrasing sounds like we hard fail and never check the resource again (even though you say something else on 143). Would like to make this a bit more clear here.

reference/managed-delivery/resource-status/index.md
emjburns: I agree with Luis - they should be in the order you're likely to see them.

emjburns (approved):
Some small wording suggestions. Otherwise, I love this and it's great and you're great. Thank you!

reference/managed-delivery/resource-status/index.md
erikmunson: 🤷‍♂ I've been thinking of it as "N" actions instead of one, because I guess theoretically we might choose to dispatch multiple tasks at a time? Either way I'm fine with this since ideally the history view will tell you about the diff -> task(s) mapping.

reference/managed-delivery/resource-status/index.md
erikmunson: Ok so I switched this to Created, Diff, Actuating, Happy, Unhappy, Paused, Resumed, Error and Unknown — paused/resume weren't accounted for in the list


d396349

reference/managed-delivery/resource-status/index.md
luispollo: Agreed as well! I didn’t think about how users would land in each section when I made that comment.

reference/managed-delivery/resource-status/index.md
robfletcher: Specifically we do launch one separate task per region for most resource types

luispollo (approved):

PR #1589, opened 2019-11-07
docs(managed): resize status images and remove useless captions

erikmunson:
These images were _way_ too big. I made them look less silly. Also got rid of the useless captions which ended up laying themselves out weirdly on the side of the images once they weren't full-width.

**Before:**
Screen Shot 2019-11-07 at 3 46 31 PM

**After:**
Screen Shot 2019-11-07 at 3 46 18 PM


emjburns (approved):
Thank you!

Issue #1597, opened 2019-11-13
Cleanup the Videos Section Success Stories Page


317brian:
The videos section should be a grid. 

https://www.spinnaker.io/success-stories/

The layout of the page could also use some TLC, e.g. collapse parts of the testimonials w/ <.details> tags.

317brian:
@stuposluns did the cleanup in https://github.com/spinnaker/spinnaker.github.io/pull/1603, so I'm closing this.

PR #1600, opened 2019-11-14
Video Grid on Success Stories Page

dnilasor:
I have implemented a simple grid in the videos area to reduce page length, in response to issue #1597 . Videos can be full-screened so the small size shouldn't be an issue. I believe this to be responsive. I won't close the issue because it mentions other page action items.

Potential next steps I see for the videos area are:

- record more vendor-neutral videos at Spinnaker Summit
- access YouTube channel to add new videos and migrate existing videos to neutral space
- update video embeds to refer to Spinnaker channel versions where applicable
- add link to YouTube Channel to bottom of the videos-container
- add any new (or old) videos we like and want showcased to the page
- add more sophisticated CSS to divide video labels from videos (give labels their own grid spaces, perhaps in a different color?) I think more flexibility in title length would be helpful
- implement the videos programmatically so they can be iterated through in index.html instead of repeating tags etc. 

Thanks for looking!

Rosalind


dorbin (approved):
LGTM, thanks Rosalind

317brian:
Awesome job!

PR #1604, opened 2019-11-19
docs(setup/security/authorization): Fix Admin API Url

progrhyme:
The old url leads us to the error page which says `API solution not found with service name: admin.`

I do not know it **was** correct url, but I think it should be fixed.


dorbin (approved):
Thanks @progrhyme

PR #1607, opened 2019-11-20
docs(secrets): Added doc for GCS secrets

german-muzquiz:
Added instructions for using GCS secrets engine released in `1.17`.

### DIFF
--- a/reference/halyard/secrets/gcs-secrets.md
+++ b/reference/halyard/secrets/gcs-secrets.md
@@ -28,7 +28,7 @@ github:
   token: <TOKEN>
 ```
 
-Note: You could choose to store the password under different keys than `github.password` and `github.token`. You'd just need to change how to reference the secret further down.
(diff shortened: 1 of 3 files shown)


reference/halyard/secrets/gcs-secrets.md
```diff
@@ -0,0 +1,56 @@
...
+### Storing credentials
+Store your GitHub credentials in `mybucket/spinnaker-secrets.yml`:
+
+```yaml
+github:
+  password: <PASSWORD>
+  token: <TOKEN>
+```
+
+Note: You could choose to store the password under different keys than `github.password` and `github.token`. You'd just need to change how to reference the secret further down.
```
dorbin:
```suggestion
Note: You could choose to store the password under different keys than `github.password` and `github.token`. You'd just need to [change how to reference the secret](#referencing-secrets).
```

dorbin (approved):
One small suggestion, @german-muzquiz . LGTM otherwise.

PR #1608, opened 2019-11-21
docs(setup/spin): Fix Url of OAuth setup

progrhyme:
I found the old url leads us towards "Page Not Found" now.


dorbin (approved):
Thanks, @progrhyme. Good catch!

PR #1609, opened 2019-11-22
add additional note regarding waiting before connecting!

MichaelDBrown:
It was not clear to me that even though halyard had completed successfully that all of the containers were not in a ready state. Until they are, using the hal deploy connect command and connecting to localhost:9000 will not work as connections will be refused. I added some wording in a note to suggest either watching the containers go ready or just waiting. In my case in a GCP deployment, 9 minutes needed to pass before all containers were ready and connections were successful. I'm open to other wording suggestions, but I do feel that a note here would be helpful to first time installers.

### DIFF
--- a/setup/install/deploy.md
+++ b/setup/install/deploy.md
@@ -55,7 +55,7 @@ deploy apply`.
 
 
 __Note:__ Even if the `hal deploy apply` command returns successfully, the 
-installation may not be complete. This is especially the case with 
+installation may not be complete yet. This is especially the case with 
 kubernetes distributed installs. If you see errors such as `Connection refused`
 it may be that all of the containers are not yet available. You can either wait, 
 or check the status of all of the containers using the commands for your cloud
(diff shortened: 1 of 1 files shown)


setup/install/deploy.md
```diff
@@ -52,6 +53,15 @@ deploy apply`.
 
 1. Navigate to [localhost:9000](localhost:9000).
 
+
+__Note:__ Even if the `hal deploy apply` command returns successfully, the 
+installation may not be complete. This is especially the case with
```
dorbin:
```suggestion
installation may not be complete yet. This is especially the case with 
```

dorbin: My thinking here is that adding this one word makes it clear that it's a waiting thing, rather than something just didn't finish.

MichaelDBrown:
Accepted @dorbin suggested edit

dorbin (approved):

PR #1615, opened 2019-12-03
update architecture docs; gate communicates with igor

anthonywoo:
I believe Gate communicates with Igor

In the Architecture > System dependencies diagram, it does not show any connection between Igor and Gate https://www.spinnaker.io/reference/architecture/#system-dependencies


duftler (approved):
LGTM.

PR #1617, opened 2019-12-03
chore(releases): Deprecate recently updated releases

ttomsu:
1.15 is currently running, but the other two are out. I'll update when 1.15.7 is out, and then we can merge this.


dorbin (approved):

ttomsu:
1.15.7 is now out too.

PR #1619, opened 2019-12-03
fix(plugins): Upgrade Jekyll plugins, and fix new errors

ttomsu:
Upgrades in `minimal-mistakes-jekyll.gemspec`:
- `bundler` to 2.0 in order to make it build (this may be why it's currently not building, not 100% sure)
- `nokogiri` to fix vulnerability warning from GH
- Unlock `jekyll-redirect` a bit to fix some generated warnings.

This made `bundle install` happy - it regenerates the `Gemfile.lock` file with new dependencies.

Other changes when testing page generation locally:
- _config.yml gems to plugins (fixed warning)
- oracle codelab had Liquid (page generation library) syntax errors because it has an embedded Packer file with conflicting syntax


ttomsu:
Note this generates a new warning:
```
NOTE: Inheriting Faraday::Error::ClientError is deprecated; use Faraday::ClientError instead. It will be removed in or after version 1.0
Faraday::Error::ClientError.inherited called from /usr/local/google/home/ttomsu/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/octokit-4.14.0/lib/octokit/middleware/follow_redirects.rb:14.
```

I'm not sure how to fix this, but I don't think it hinders page rendering

dorbin (approved):

PR #1621, opened 2019-12-03
fix(docs): Rely on Github managed gem for dependencies

ttomsu:
Following these instructions https://jekyllrb.com/docs/github-pages/

I hope this change ensures we always stay compatible with the current Github Pages [dependency versions](https://pages.github.com/versions/)


dorbin (approved):
LG as far as I can tell. Let's see what happens.

PR #1627, opened 2019-12-10
docs(stats): Adds initial stats collection description page

ttomsu:
@bdmorgan

Initial cut - probably needs some more detailed and persuasive language.

### DIFF
--- a/community/stats/index.md
+++ b/community/stats/index.md
@@ -1,0 +1,66 @@
+---
+layout: single
+title:  "Usage Statistics"
+sidebar:
+  nav: community
+---
+
+{% include toc %}
+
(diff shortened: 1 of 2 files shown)


community/stats/index.md
```diff
@@ -0,0 +1,64 @@
...
+layout: single
+title:  "Usage Statistics"
+sidebar:
+  nav: community
+---
+
+{% include toc %}
+
+## Why is data being collected?
+
```
dorbin, line 13: Maybe a sentence here, without a heading.

"[If enabled], Spinnaker collects data about how the tool is being used. This data is anonymized and aggregated.

Or words to that effect. "If enabled" would link to the thing you have below about opt-in/opt-out.

dorbin (approved):

community/stats/index.md
ttomsu: Done.

PR #1646, opened 2019-12-18
Task/add cdf ga

danlopez00:
Add CDF GA and analytics provider include


ttomsu:
This looks like it's in addition to the one we already have? Do we need two or can we share the one we have?

ttomsu:
Ping

danlopez00:
We are trying to consolidate metrics across all CDF projects. So we have two options

- Add CDF team members to your GA account, or 
- start using CDF CA sub account

Open to both scenarios

plumpy:
It looks like you can move a property from one Google account to another: https://support.google.com/analytics/answer/6370521?hl=en

Would it work to just move our current property under your account (which would also preserve old data), or are you trying to consolidate all CDF-owned websites under a single property?

PR #1650, opened 2019-12-19
Fix pipeline template guide

neakor:
```bash
spin pipeline-templates plan --file
```
The parameter should be referring to the pipeline template file not the pipeline file.


louisjimenez (commented):
Are you seeing an error when using `spin pipeline-templates plan --file <file.json>` with a pipeline config arg? The pt plan command is intended to give a view of the fully hydrated pipeline config (with stages, resolved variables and inherited items) once the template reference has been resolved. It accepts a pipeline config so that it can follow the template reference and return the hydrated pipeline. If you are able to share a redacted config and template, I can try to reproduce what you are seeing.

neakor:
This is only fixing the verbiage of the guide. The actual `spin` CLI is just fine. There are no errors.

louisjimenez:
Happy to hear that there are not any errors. However, this change would lead users to believe they should provide a pipeline config as the argument to the spin pt plan command. Take a look at the source for the Spin CLI, it indicates that a template is the desired arg for this command, https://github.com/spinnaker/spin/blob/master/cmd/pipeline-template/plan.go#L49

neakor:
I think that's exactly what I'm suggesting here. It used to say
`spin pipeline-templates plan --file <path to pipeline config>`
I'm suggesting to change it to
`spin pipeline-templates plan --file <path to pipeline template file>`

louisjimenez:
Oof, my mistake, I wrote the last response a little too quickly and swapped the two in my previous comment. My original comment before that had it correct, the intended argument is a pipeline config. That lets Spin CLI provide the pipeline config to execute a plan in Orca.

neakor:
Oh that's interesting. So the `plan` command takes in actual pipeline configs even though it's namespaced under `pipeline-templates`?

louisjimenez:
Yes, it expects a pipeline config. All the commands that involve the concepts of managed templates were grouped under the pipeline template namespace.

neakor:
Thanks for clarifying.

PR #1655, opened 2019-12-23
Update index.md

alonm-totango:
Page: [/concepts/](https://www.spinnaker.io/concepts/)

Feedback:
I can't stress enough how important it is to build up from the smallest component in an explanation rather than tare down the biggest concept.
Starting with the application and using every concept that hasn't been explained yet is very confusing and unsatisfying. 

I would start by rearranging the order of the concepts from the bottom up:
1) Firewall
2) Load Balancer
3) Server Group
4) Cluster
5) Application


317brian:
What I'd like to see more of on the page is tying each thing into how it fits into the Spinnaker big picture/gets deployed by Spinnaker.

I think most people know what a firewall is, but it's missing that second part. 

With that said, I'm not opposed to trying out this alternative organization.

robzienert:
@alonm-totango Thanks for the contribution! Looks like there's some feedback here.

PR #1658, opened 2020-01-06
docs(managed-delivery): Describe support for git-based workflows

luispollo:
Add basic documentation for git-based workflow support in Managed Delivery.


reference/managed-delivery/gitops.md
```diff
@@ -0,0 +1,23 @@
+---
+layout: single
+title:  "Managed Delivery"
+sidebar:
+  nav: reference
+---
+
+{% include toc %}
+
+Spinnaker supports _GitOps_ workflows to create and update Managed Delivery configurations.
```
erikmunson: Thoughts on switching to more simplistic language than `GitOps`, e.g. `Git integration` and `Git-based workflows`? GitOps can be a pretty loaded term depending on individual background and experience.

reference/managed-delivery/gitops.md
```diff
@@ -0,0 +1,23 @@
...
+ 
+GitOps support is currently achieved via a small pipeline configured with a git
+[trigger](../../guides/user/pipeline/triggers/index.md) and containing a single
+`Import Delivery Config` stage. This stage will retrieve a Delivery Config manifest
+from the source control repository associated with your pipeline's trigger, then save (or update)
+it in Spinnaker so it will automatically monitor and manage the resources, environments and
+deployments described in the manifest.
+
+Eventually, we might add support for direct monitoring of git commit events to Managed Delivery
+so that you don't even need a pipeline, but for now, this is the way to do it.
```
erikmunson: this is a very @emjburns sentence 👍

erikmunson:
Lovely. I suggested a wording tweak but overall text looks good. Only other thing that might be helpful is a screenshot (or multiple) of the stage to help ground people in a reality a bit more when they're reading it.

reference/managed-delivery/gitops.md
luispollo: That's an excellent question... I really don't know. On the one hand, I feel like, for as much as I agree with you on the term being overloaded, most people in the industry would have a general understanding of what "GitOps" means here and how it applies; on the other, I don't think we necessarily gain too much from using it other than buzzword compliance. Another alternative would be to make that a link to one of the more common articles on the topic (like [this one](https://www.weave.works/blog/what-is-gitops-really) from WeaveWorks, who coined the term).

luispollo:
> Lovely. I suggested a wording tweak but overall text looks good. Only other thing that might be helpful is a screenshot (or multiple) of the stage to help ground people in a reality a bit more when they're reading it.

Thanks @erikmunson! I thought about adding screenshots but was afraid that https://github.com/spinnaker/deck/pull/7733 might still change as a result of review. Will keep this pending and update with screenshots when that one is ready to merge.

reference/managed-delivery/gitops.md
erikmunson: I think seeing that WeaveWorks article further reinforces my gut feel that GitOps has too much conceptual baggage — anecdotally when I've used the term I've noticed folks assume a bunch of things (e.g. you're on K8s, or the definition is statically all there rather than having some dynamism, or you commit changes to spec files to kick off software deploys, etc etc). Generally I would say that Netflix and Managed Delivery don't really agree with the 'strict' definition of GitOps, as we would like software versions and other types of details to change out-of-band from version control.

luispollo: Fair points. I removed the references to GitOps.

reference/managed-delivery/gitops.md
```diff
@@ -0,0 +1,23 @@
...
+---
+
+{% include toc %}
+
+Spinnaker supports git-based workflows to create and update Managed Delivery configurations.
+This allows you to keep your Managed Delivery configuration in code and follow all the usual
+best practices like code review and approvals before a change is introduced to your managed
+resources, environments or deployments.
+ 
+Git support is currently achieved via a small pipeline configured with a git
```
emjburns:
```suggestion
Git support is currently achieved via a pipeline configured with a git
```

(small seems unnecessary here)

emjburns (approved):

reference/managed-delivery/gitops.md
317brian: If we're dropping `GitOps` from the text, should the file get renamed too?

317brian:
For this to show up in the navigation on https://www.spinnaker.io/reference/, you need to add an entry to the `reference` section of `_data/navigation.yml` for this and the other Managed Delivery pages.

reference/managed-delivery/gitops.md
luispollo: Will do.

luispollo:
> For this to show up in the navigation on https://www.spinnaker.io/reference/, you need to add an entry to the `reference` section of `_data/navigation.yml` for this and the other Managed Delivery pages.

Thanks @brian-armory. We're intentionally making these pages non-discoverable for now.

317brian:
Sounds good!

PR #1659, opened 2020-01-09
Update Service Account Role

SimonMuraya:
Service Account Actor role
This role has been deprecated. If you need to run operations as the service account, use the Service Account User role. To effectively provide the same permissions as Service Account Actor, you should also grant Service Account Token Creator.
Ref: https://cloud.google.com/iam/docs/service-accounts#the_service_account_actor_role


dorbin (commented):
Do we need to include these decriptions, including the notification the the actor role is deprecated? Or should we just replace the command with the new one that includes the serviceAccountUser role?

dorbin:
Adding @ezimanyi for confirmation that this is the role to use.

ezimanyi:
Seems reasonable to me---I didn't realize that "Service Account Actor" had been deprecated, but based on the linked docs it looks like the updated version should grand the equivalent permissions using non-deprecated roles.

dorbin (approved):

PR #1660, opened 2020-01-10
fix(crd): update CRD documentation

ethanfrogers:
update CRD documentation to reflect changes to where the referenced code lives.


ezimanyi (approved):
Good catch, thanks for fixing this!

PR #1664, opened 2020-01-13
docs(plugins): added docs on frontend development for plugins

bpowell:


christopherthielen:
In the short term this documentation looks good.  Medium term, I'd like to work towards a [pre-scaffolded plugin repo](https://github.com/bpowell/community/blob/rfc-ui-plugins/rfc/plugins-ui.md#plugin-development) that developers can clone.  I'd also like most of the opinions (such as rollup config and provided libraries) to be encapsulated in the `@spinnaker/plugin` dependency so those opinions can be updated over time.

christopherthielen (approved):

aimeeu:
Thanks for the update @bpowell. lgtm from a writing standpoint.

christopherthielen:
There is now scaffolding available via this command `npx -p @spinnaker/pluginsdk scaffold`

PR #1671, opened 2020-01-16
docs(java11): add note about java 11 migration

317brian:
We discussed adding this note during the SIG meeting before the holidays.


dorbin (approved):

plumpy:
It's perfect, thanks!

PR #1673, opened 2020-01-21
docs(halyard): fix format issue in halyard.md

burov:


317brian (approved):

317brian:
Thanks for catching this! We try to use the same PR title format as the code repos, so I went ahead and updated that for you.

PR #1675, opened 2020-01-21
fix(crd): fix crd extension docs

ethanfrogers:
the code in the `KubernetesHandler` section was outdated and didn't work
with an extension. I've updated the method signatures and content to
reflect the state of the class today.

### DIFF
--- a/guides/developer/crd-extensions.md
+++ b/guides/developer/crd-extensions.md
@@ -46,7 +46,7 @@ public class MyCRDHandler extends KubernetesHandler {
 
   @Override
   public KubernetesKind kind() {
-    return "MyCRDKind";
+    return KubernetesKind.from("MyCRDKind");
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


guides/developer/crd-extensions.md
```diff
@@ -70,10 +70,8 @@ public class MyCRDHandler extends KubernetesHandler {
   }
 
   @Override
-  public Class<? extends KubernetesV2CachingAgent> cachingAgentClass() {
-    // Caching agent class for your CRD.
-    // See, e.g., `KubernetesReplicaSetCachingAgent`.
-    return MyCRDCachingAgent.class;
+  public KubernetesV2CachingAgentFactory cachingAgentFactory() {
+    return KubernetesCoreCachingAgent::new;
```
ezimanyi, line 74: Do we maybe want this to be `MyCRDCachingAgent::new` to map onto what was there before?

ezimanyi (approved):

guides/developer/crd-extensions.md
ethanfrogers: I'm not sure, actually. Do you actually need a custom caching agent for this? When you configure `customResources`, we use the `CustomKubernetesHandlerFactory` to generate one. Maybe we should just do that?

ezimanyi: I presumed that if people were actually extending clouddriver, it's because they wanted to wrtie their own caching agent (rather than just register additional CRDs via config)---but maybe there's a reason I'm missing for writing this class but re-using an existing caching agent?

ethanfrogers: The primary use case we're working against right now is adding stability logic. Based on the examples I pulled from, it looks like those who are doing this today are just using something like the Core caching agent or `KubernetesUnregisteredCustomResourceCachingAgent::new`.

ezimanyi: Oh, that makes sense, yeah in that case I think we can leave one of the existing caching agents in here.

PR #1688, opened 2020-01-28
Update definition of stage and add definition of task

fdegir:
This PR is a result of a discussion which took place under
another PR. [0] Sending this PR in order to update the Spinnaker
documentation to ensure people who read the documentation can
get the correct information about stages and tasks.

[0] https://github.com/cdfoundation/sig-interoperability/pull/14#discussion_r371995711


concepts/index.md
```diff
@@ -101,14 +101,18 @@ pipeline start/complete/fail).
 
 ### Stage
 
-A *Stage* in Spinnaker is an atomic building block for a pipeline, describing an
-action that the pipeline will perform. You can sequence
-[stages](/reference/pipeline/stages/) in a Pipeline in any order, though some
+A *Stage* in Spinnaker is a collection of sequential Tasks and composed Stages that
+describe a higher-level action the Pipeline will perform either linearly or in parallel.
```
dorbin, line 105: Higher level than a task, for sure, but what distinguishes a task from a stage, conceptually? That is, I have this thing that needs to be done as part of my pipeline; how do I know if it's a stage or a task in a stage?

concepts/index.md
```diff
@@ -101,14 +101,18 @@ pipeline start/complete/fail).
 
 ### Stage
 
-A *Stage* in Spinnaker is an atomic building block for a pipeline, describing an
-action that the pipeline will perform. You can sequence
-[stages](/reference/pipeline/stages/) in a Pipeline in any order, though some
+A *Stage* in Spinnaker is a collection of sequential Tasks and composed Stages that
```
dorbin, line 104: What do we mean by "composed stages" making up a stage?

concepts/index.md
fdegir: I'm not very into the details as much unfortunately so I'll need help from @robzienert and others to explain these terms.

concepts/index.md
fdegir: ditto

concepts/index.md
robzienert: From https://www.spinnaker.io/guides/developer/service-overviews/orca/#domain

> Within a Stage, there are one or more Tasks (not to be confused with the Tasks tab in the UI). A Task can be considered an atomic unit of work, focused on a single action. If you are familiar with the Spinnaker UI, you’ll note that for every Stage, there are multiple line items that correlate to a Stage’s runtime: These are often Tasks. Tasks are always executed sequentially and serially.
> 
> It’s important to note that a Stage is recursively composable. One stage can have zero to many Synthetic Stages, that is, stages that can occur either BEFORE or AFTER its parent stage. This is most easily illustrated in a Canary stage, which deploys multiple Server Groups (a baseline and canary). While to the end-user this appears to be a single stage, it’s actually a composition of multiple Deploy stages with some extra logic on top.
> 
> Lastly, an Execution’s stage graph does not need to be known ahead of time. An Execution can lay its tracks down ahead of itself as it is running, which is how some of the more advanced functionality is implemented, like automatic rollbacks or canary behaviors.

fdegir:
just wondering what to do with this PR? I still think it is important to update the documentation in order for people to get correct information about terms used in Spinnaker.

I'm fine to close it if someone else has better explanation of the terms and can do that.

robzienert:
Will merge and loop back on changes.

PR #1695, opened 2020-01-31
chore(logos): Adding Cure.fit and Diligent Success Story Logos

stuposluns:


317brian (approved):
Thanks Stu! I fixed some errant commas and replaced the : w/ the character code for a colon.

317brian:
Investigating why the rake test fails. My local copy of the site builds and renders fine.

317brian:
Solved with properly escaped and character codes for special characters

317brian (approved):
Yay the rake test passed

PR #1704, opened 2020-02-11
docs(notices): remove old notices

317brian:
Removes the 1 year+ version notices that use `include alpha version`. Will submit subsequent ones based on wider searches. 

The K8s v2 notice is less than 1 year old now, but if we're deprecating v1, it makes sense to me to remove the alpha from v2.


dorbin (approved):
Yeah, I don't think there's a mechanism in place now to expire/promote alpha features. But you're right about v2 docs. I did a quick spot-check just now, and I don't see any other v2 pages with alpha notices.

PR #1710, opened 2020-02-18
chore(kubernetes): highlight upcoming v1 provider removal

maggieneterval:
- chore(kubernetes): link to V1 provider removal RFC

  Replace "maintenance mode" with "removal" and link to the [spinnaker/governance RFC](https://github.com/spinnaker/governance/blob/master/rfc/eol_kubernetes_v1.md). I will adjust the language here to reflect the fact that the V1 provider _was removed_ when we release 1.21.

- chore(kubernetes): replace link to clouddriver at master with hash

  In the future, let's try to avoid linking from documentation to code. Since this particular link was useful enough that [a community member asked for it to be fixed](https://github.com/spinnaker/spinnaker/issues/5458), let's pin it to a specific commit hash rather than `master`.

### DIFF
--- a/reference/providers/kubernetes.md
+++ b/reference/providers/kubernetes.md
@@ -6,6 +6,9 @@ sidebar:
 redirect_from: /reference/providers/kubernetes-v1/
 ---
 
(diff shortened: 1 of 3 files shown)


ezimanyi (approved):
:memo: :arrow_up: :100:

dorbin (approved):
:+1:

PR #1718, opened 2020-02-25
Fix typo

ejsuncy:


317brian (approved):

317brian:
Thanks for catching that!

ejsuncy:
I mean I’m all about shots but...

PR #1720, opened 2020-02-25
Provide a link to field descriptions carried over from v1 schema

ejsuncy:


reference/pipeline/templates/index.md
```diff
@@ -80,3 +80,5 @@ config:
   "stages": []
 }
 ```
+
+A more thorough explanation of some fields such as `protect` can be found in the `v1` schema [here](https://github.com/spinnaker/dcd-spec/blob/master/PIPELINE_TEMPLATES.md).
```
dorbin, line 84: Are some fields not included in the linked doc?

ejsuncy: When following the [guide for creating a pipeline template from a pipeline](https://www.spinnaker.io/guides/user/pipeline/pipeline-templates/create/#3-edit-the-file-for-template-format) it took me a long time to find what the `protect` field is for. After googling, I found the [Pipeline Templates reference spec](https://www.spinnaker.io/reference/pipeline/templates/) which merely shows it as ` "protect": <true | false>` with no other comments. I had to google a while for spec v1 (which is why I link to it here), which shows a more descriptive `protect: A flag (defaults false) to control whether or not configurations can change or mutate the template's defined stage graph at plan/exec time. Side effects of template variables are not included in this protection.` I'm just trying to save time for others that are trying to follow the guide.

dorbin (approved):
Sorry, @ejsuncy , this has languished awhile. No reason not to merge this helpful addition. Thanks!

PR #1736, opened 2020-03-08
docs(AWS/S3):Mention 'api-endpoint' & 'api-region' are optional

agentmilindu:
Explained the need for 'api-endpoint' & 'api-region' flags are optional and only if using Minio.

I'm a beginner to Spinnaker. I wondered what it means by  'api-endpoint' & 'api-region'.
Only after reading more options link I realized this is optional. Thus mentioning to avoid confusions for beginners.


setup/artifacts/s3.md
```diff
@@ -7,7 +7,7 @@ sidebar:
 
 {% include toc %}
 
-Spinnaker stages that read data from artifacts can read S3 files directly.
+Spinnaker stages that read data from artifacts can read S3 files directly. Setting `--api-endpoint` & `--api-region` is optional. It is needed only if you are using a S3 clone such as Minio.
```
allisaurus: What do you think about moving this to below the code snippet in  "2. Add an Artifact to your account"? 
We're introduced to these flags there, so it might flow better to make mention of them there like: "Note: setting `--api-endpoint` and..."

agentmilindu: Make sense. I'll move it there :)

allisaurus (approved):
LGTM!

317brian (approved):
@allisaurus thanks for the confirmation!

PR #1746, opened 2020-03-16
chore(release): Deprecate 1.16.7

ezimanyi:
As we support 3 minor releases, the release of 1.19.0 means that the 1.16.x releases are now deprecated.


maggieneterval (approved):
👋 1️⃣ 6️⃣

PR #1747, opened 2020-03-18
refactor(changelog): Update changelog links

ezimanyi:
To simplify the build process, rather than create an entirely new gist for every release, we'll only create a gist for each minor Spinnaker version (1.18.x, 1.19.x, etc.). That gist will contain the release notes for each patch release within that minor version as separate files (ex: 1.18.2.md, 1.19.1.md).

I've created these gists for the 1.17, 1.18 and 1.19 branches; this points the docs site to these new gists. By passing the file URL parameter we will only embed the specific file related to the patch release in question, which means this change will result in no difference from a user's perspective; it just simplifies creating these gists.

### DIFF
--- a/_changelogs/1.17.0-changelog.md
+++ b/_changelogs/1.17.0-changelog.md
@@ -5,4 +5,4 @@ date: 2019-11-04 14:07:43
 tags: changelogs 1.17 deprecated
 version: 1.17.0
 ---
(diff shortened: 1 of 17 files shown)


maggieneterval (approved):
♻️ 📜 🔤 👉 🗄 📁 💯

PR #1748, opened 2020-03-18
refactor(changelog): Revert changelog links

ezimanyi:
It turns out there's a problem with using a single gist for all of the changelogs in a minor version; the maximum size that will be rendered for a gist is across all of the files in a gist, and the 1.17.7 release pushed the 1.17.x gist across that size limit.  This means that trying to view the 1.17.7 release notes now just yields a message that the gist can't be rendered and a link to view the raw gist.

Reverting this as we'll need a different approach (maybe that involves less copy-pasting of the content so the gist doesn't grow as N^2 in the number of patch releases).

This reverts commit b8084fbba1b842e2c6d4b6720ba9aa0f3ab17fbc.

### DIFF
--- a/_changelogs/1.17.0-changelog.md
+++ b/_changelogs/1.17.0-changelog.md
@@ -5,4 +5,4 @@ date: 2019-11-04 14:07:43
 tags: changelogs 1.17 deprecated
 version: 1.17.0
 ---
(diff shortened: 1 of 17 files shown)


maggieneterval (approved):
04B9DAC3-DAE7-4709-8F9D-3A64527CFBEE

PR #1749, opened 2020-03-18
chore(build): Simplify changelog gists

ezimanyi:
The prior attempt to simplify changelog gists tried putting all the changelogs for a particular minor release as different files in the same gist.  This failed because for some releases we exceed the size that can be rendered for a gist (which causes failed rendering even if the particular file we're viewing is less than the max size).

The root cause was that we were copy-pasting the prior patch release's changlog into the new changelog and appending changes at the top.  This mean that, for example for 1.17.x, with 7 patch releases, we had the 1.17.0 release notes copied 7 times.

Rather than manually copy these release notes from gist to gist, let's just update the website to display the gists for the current patch release and all prior patch releases in order. This means that when making a new patch release, the gist just needs to contain the notes for that patch; the site will then incorporate the prior notes by reference. I have a corresponding update to the buildtool to update how we generate these changelog files, so they automatically pull all patch releases, but this PR is to backfill the currently-supported releases.  (I actually used the buildtool locally to generate this change.)

With this change, we greatly reduce the total size of the gist for each release, getting around the earlier problem.

### DIFF
--- a/_changelogs/1.17.1-changelog.md
+++ b/_changelogs/1.17.1-changelog.md
@@ -6,3 +6,4 @@ tags: changelogs 1.17 deprecated
 version: 1.17.1
 ---
 <script src="https://gist.github.com/spinnaker-release/d020714e9190763f27e35701e14c6bc1.js?file=1.17.1.md"/>
+<script src="https://gist.github.com/spinnaker-release/d020714e9190763f27e35701e14c6bc1.js?file=1.17.0.md"/>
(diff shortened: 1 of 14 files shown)


ezimanyi:
Also, while working on this, I noticed that somehow my revert PR #1748 actually contained no code...I'm not sure how that happened and how I didn't notice 😬 , but this actually fixes the issue 😀

maggieneterval (approved):
📜 3️⃣ 👉 📜 2️⃣ 👉 📜 1️⃣ 👉 📜 0️⃣

PR #1750, opened 2020-03-18
fix(build): Fix script tags in changelog files

ezimanyi:
For some reason, only the first changelog is showing up after the last commit. Digging into this, I believe it's because we are incorrectly using a self-closing tag for the script element. This seems to have been working for a single script element but is breaking when there is more than one, causing only the first to show up.


maggieneterval (approved):
🤳 🚪 🔐 🏷

PR #1752, opened 2020-03-19
fix(docs): update required node version

maggieneterval:
The required minimum `node` version is now 12.16.0. I was torn between updating its value here, and changing these instructions to link to deck's [package.json](https://github.com/spinnaker/deck/blob/master/package.json) and instructing users to install the version number at `engines.node`, but given the relative infrequency of minimum `node` version updates, let's just update the value for now.


ezimanyi (approved):
🆙 📆 ☕️ 📜 ✅

PR #1754, opened 2020-03-21
chore(update): Home Page Promos

dnilasor:
`In this PR, I've updated the conference ad block on the home page to feature information and links for Spinnaker Gardening Days. I've also added a simple Join Slack CTA under that conference ad block, with the slack logo. Ideally this block would be a bit narrower so that it doesn't push other page content down so much, but to me this is acceptable in the short-term while we refresh the site. My reasoning for adding the CTA is that, I'm worried folks are coming to the site home page now and not knowing where to turn for help. We can help OSS users much better on Slack, so they don't get stuck and frustrated on their own.
@brian-armory @dorbin @aimeeu take a peek if you like.


317brian:
LGTM. One question before I approve and merge, is the image public domain or some other free use license?

index.md
```diff
@@ -1,10 +1,15 @@
 ---
 layout: home
 conference_ad_row:
```
dorbin: @dnilasor do you mean to keep the previous summit-row name? No big deal if you don't change it.

dorbin (approved):
LGTM, with one small comment. 
Thanks!

index.md
dnilasor: I did because it seemed like a generic-enough row name. Maybe I should change it to an event more generic name (promo_row ?) so we don't need to change it? @dorbin maybe there is some value there

dnilasor:
I added another change with a descriptive message. Please LMK if anything else needed before approval!

index.md
```diff
@@ -1,12 +1,17 @@
...
-  image_path: assets/images/summit-socialmedia-transparent.png
-  alt: "Spinnaker Summit Logo"
+promo_row:
+  title: Sign Up for Virtual Spinnaker Gardening Days
+  excerpt: "Have fun and move the needle on continuous delivery at Spinnaker Gardening Days #communityHack, sponsored by Salesforce! [Register](https://go.armory.io/gardening) and join Spinnaker users and companies on April 9 & 10 to learn and build. Enjoy a collaborative hackathon, free training, and community celebration. We’ll dig into open-source issues and features as well as extend the ecosystem. [Click to learn more](https://github.com/spinnaker-hackathon/gardening)."
+  image_path: assets/images/gardening.png
+  alt: "Spinnaker Gardening Days Logo"
+slack_row:
+  title: Join Spinnaker Slack
+  excerpt: "Visit [join.spinnaker.io](https://go.armory.io/spinslack) to get help on Spinnaker Slack, a place for the community to come together. Use this vibrant workspace to ask and answer questions, connect with other operators and users, discuss issues with SIGs, and learn Spinnaker!"
```
dorbin: Hm, might be better to link directly to join.spinnaker.io rather than to an Armory url that redirects there?

index.md
```diff
@@ -1,12 +1,17 @@
...
+  image_path: assets/images/slack_logo.png
+  alt: "Join Spinnaker Slack"
 spinnaker_row:
-  title: Spinnaker is an open source, multi-cloud continuous delivery platform for releasing software changes with high velocity and confidence.
+  title: Spinnaker is an open source, multi-cloud continuous delivery platform for releasing software changes with high velocity and nczidence.
```
dorbin: Inadvertent?

index.md
```diff
@@ -1,12 +1,17 @@
 ---
 layout: home
-conference_ad_row:
-  title: Spinnaker Summit 2019
-  excerpt: "Spinnaker Summit 2019 took place in San Diego, CA from November 15-17th. Hundreds of people from the Spinnaker community gathered and shared stories about their experiences with Spinnaker and Continuous Delivery. See the insights from Spinnaker contributors and users in [full session recordings here](https://www.youtube.com/playlist?list=PL4yLrwUObNkvGzXKSEU8miQTlvZyZvXWq)."
-  image_path: assets/images/summit-socialmedia-transparent.png
-  alt: "Spinnaker Summit Logo"
+promo_row:
```
dorbin, line 3: Yeah, this!

index.md
dorbin:
```suggestion
  excerpt: "Visit [join.spinnaker.io](https://go.armory.io/spinslack) to get help on Spinnaker Slack, a place for the community to come together. Use this vibrant workspace to ask and answer questions, connect with other operators and users, discuss issues with SIGs, and learn about Spinnaker!"
```

dorbin: Just because "..and learn Spinnaker" almost implies Spinnaker Slack is sufficient for learning Spinnaker.

index.md
dnilasor: Yes, wth oops

dnilasor: fixed

index.md
dnilasor: @dorbin understood, here's my commit message on this:

>  I received a request that we be able to track how many folks join slack using the link from spinnaker.io. The easiest way for me to do this currently is to use Armory's existing rebrandly account which includes detailed link tracking, so I've added that URL here. I'm sure there is a way to do this with Google Analytics and Tag Manager, although I don't know it off the top of my head. I want to know, are we comfortable with me using the rebrandly approach if we're not using it for anything other than tracking? Or do I need to look into an alternative approach? Thanks!

I'd love your thoughts @dorbin @brian-armory !

dnilasor:
@plumpy, makes sense. Thanks for putting #1646 on my radar. Seems like we need to sort out implementation details, so for now I'll just re-add the simple join.spinnaker.io link so we can get this approved. Then I'll look for the most maintainable strategy. Thanks for looking @dorbin @plumpy !!

dnilasor:
@brian-armory, changed the link to something not controversial and will open another PR that addresses this and #1646. Mergey merge time please?

dorbin (approved):

317brian:
Thanks for making sure the analytics discussion doesn't get lost @dnilasor. I'll go ahead and merge since Dave already approved.

317brian:
Weird. What appears on the OSS site is not what showed locally when I built it originally. @dnilasor will troubleshoot.

PR #1760, opened 2020-03-25
fix(promo): work around the css issues

dnilasor:
Hi! I think that when we changed the`conference_ad_row` to  `promo_row` we broke the CSS. My bad, I didn't test the visual thoroughly. I have reverted to `conference_ad_row` and I'm seeing it looking better in my local preview now. I don't think it is worth it to spend much time investigating CSS issues and worrying too much about naming right before the refresh!


317brian (approved):
Thanks for fixing this. It seems like the row name is the culprit. I built it locally, and it looks as expected.

dorbin:
👍

PR #1762, opened 2020-03-26
docs(cherrypick): use mergify instead of spinnakerbot

plumpy:


ezimanyi (approved):
🍒 ⛏ 
Thanks for doing this!

PR #1775, opened 2020-04-07
Docs(reference): add some more ports

dnilasor:
Add more ports to the port mapping doc. Linking to this doc from hackathon dev env doc and want to have all the non-proprietary ports I can find in our internal docs represented here.


reference/architecture/index.md
```diff
@@ -128,8 +128,11 @@ By default Spinnaker binds ports according to the following table
 | Fiat        | 7003              |
 | Front50     | 8080              |
 | Gate        | 8084              |
+| Gate x509   | 8085              |
```
ezimanyi, line 131: Is this saying that when you use x509 authentication for Gate it instead listens on port 8085?  I'm using this and it's still listening in 8084...when does it listen on 8085?

reference/architecture/index.md
```diff
@@ -128,8 +128,11 @@ By default Spinnaker binds ports according to the following table
...
 | Halyard     | 8064              |
 | Igor        | 8088              |
 | Kayenta     | 8090              |
 | Orca        | 8083              |
 | Rosco       | 8087              |
+| Redis       | 6379              |
```
ezimanyi, line 137: Would it make sense to pull these into their own section, as we're really documenting defaults of external services we don't control here?  It does seem useful to have the ports they listen on handy as you often need to run commands against them to check the state of data, but maybe as an "external services" section?

reference/architecture/index.md
```diff
@@ -128,8 +128,11 @@ By default Spinnaker binds ports according to the following table
...
+| MariaDB     | 3306              |
```
ezimanyi, line 138: Should this reference `MySQL` instead?  From [the docs](https://www.spinnaker.io/setup/productionize/persistence/orca-sql/#mariadb) it looks like it's a pretty specific use case to use the MariaDB connector to connect to AWS and most people will just be using a MySQL installation.

ezimanyi (commented):
Just a few questions, thanks!

reference/architecture/index.md
dnilasor: I like this idea!! Will add these edits soon thx

reference/architecture/index.md
dnilasor: Need to check with SME on this as I haven't enabled x509 and simply based this on internal doc. Stay tuned : )

reference/architecture/index.md
dnilasor: Yes. Could say either/or but to me MySQL and Maria are analogous so, +1. Will update

PR #1781, opened 2020-04-08
feat(md): add pinning doc

emjburns:
Pinning is available via the API. Full ui support for this is coming, but these docs cover interacting at the api-level with this feature.

### DIFF
--- a/guides/user/managed-delivery/pinning/index.md
+++ b/guides/user/managed-delivery/pinning/index.md
@@ -1,0 +1,67 @@
+---
+layout: single
+title: "Pinning an Environment"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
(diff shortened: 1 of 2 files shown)


guides/user/managed-delivery/pinning/index.md
```diff
@@ -0,0 +1,67 @@
+---
+layout: single
+title: "Pinning an Environment"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+You can pin an environment to a specific artifact version to ensure that all clusters in the environment stay on that version.
```
erikmunson:
```suggestion
You can pin an environment to a specific artifact version to ensure that all resources in the environment stay on that version.
```

erikmunson (approved):
🚀 teeny suggestion, but overall LGTM — I expect we'll flesh this out much further once the UI can handle it.

PR #1782, opened 2020-04-09
docs(plugins): Plugins Users Guide - add deploy pf4jStagePlugin example

aimeeu:
Update the Plugin Users Guide to add a page with steps on how to deploy the pf4jStagePlugin. Included are two Halyard 1.33 caveats. Without these, users will encounter errors deploying plugins.

Would like to get this merged for Gardening Days.

### DIFF
--- a/guides/user/plugins/deploy-example.md
+++ b/guides/user/plugins/deploy-example.md
@@ -31,14 +31,13 @@ These issues are fixed in Halyard 1.34.
 
 2. Create the `orca-local.yml` file in `.hal/default/profiles`. If you are using Halyard 1.33, see [Known Issues](#known-issues).
 
-3. Configure the plugin repository and redeploy Spinnaker:
+3. Configure the plugin repository:
 
 	```shell
   # Configure the plugin repository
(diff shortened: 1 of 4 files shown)


aimeeu:
@brian-armory @dorbin  I'd appreciate a review as soon as you have time.

guides/user/plugins/deploy-example.md
```diff
@@ -0,0 +1,91 @@
...
+    extensions:
+      armory.randomWaitStage:
+        id: armory.randomWaitStage
+        enabled: true
+        config:
+          defaultMaxWaitTime: 60
+	```
+
+
+1. Redeploy Spinnaker.
```
317brian: Is there a reason why they have to run `hal deploy apply` earlier in this process too? Why not have them just run it once at the end to save them some wait time?

317brian (approved):
Thanks @aimeeu. I have one question that's not critical

guides/user/plugins/deploy-example.md
aimeeu: I've never tried that  but I don't see any reason why that wouldn't work

PR #1785, opened 2020-04-09
chore(plugins): adding videos for plugin build and delivery

link108:


317brian (approved):
Thanks for getting this done @link108

PR #1794, opened 2020-04-11
fix(reference): Fix misconfiguration in the example for component sizing

dragon3:
## WHAT

This pull request fixes misconfiguration in the example for component sizing.

## WHY

The first example in `customSizing:` should be `echo`, not `spin-echo`.

### DIFF
--- a/reference/halyard/component-sizing.md
+++ b/reference/halyard/component-sizing.md
@@ -21,7 +21,7 @@ There are two ways to specify requests and limits
 
 * _(Recommended)_ By supplying the container name, e.g. `echo:`, followed by the container requests and limits.
 
-  This sets resource configuration for the echo container in `spin-echo` service's pod only, not any sidecars
+  This sets resource configuration for the `echo` container in `spin-echo` service's pod only, not any sidecars
   (e.g. the monitoring daemon).
    
 * By supplying the service name, e.g. `spin-clouddriver:`, followed by the container requests and limits.
@@ -36,7 +36,7 @@ deploymentEnvironment:
   customSizing:
     # This applies sizings to only the echo container and not to any sidecar 
     # containers running with echo.
-    spin-echo:
+    echo:
       limits:
         cpu: 250m
         memory: 512Mi


ezimanyi (commented):
@dragon3 : I think that the docs are actually correct as they are currently written.  Halyard deploys the microservices with a `spin-` prefix (ex: `spin-clouddriver`, `spin-echo`) and is also expecting that the fields in the `customSizing` map will have that prefix.

I've been using this and have mine set with the `spin-` prefix, and it's working for me.  Are you not seeing these limits applied when you configure the sizing with the `spin-` prefix?

dragon3:
@ezimanyi 
Thank you for the reply.

My fix is for this:

>     # This applies sizings to only the echo container and not to any sidecar 
>     # containers running with echo.

If we specify the service name such as `spin-echo` for customSizing like this:

```yaml
deploymentConfigurations:
  deploymentEnvironment:
    customSizing:
      spin-echo:
        limits:
          cpu: 4
          memory: 3Gi
        requests:
          cpu: 2
          memory: 3Gi
```

Halyard creates Deployment like this:

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: spin-echo
...
spec:
...
  template:
...
    spec:
      containers:
      - name: echo
        resources:
          limits:
            cpu: "4"
            memory: 3Gi
          requests:
            cpu: "2"
            memory: 3Gi
      - name: monitoring-daemon
        resources:
          limits:
            cpu: "4"
            memory: 3Gi
          requests:
            cpu: "2"
            memory: 3Gi
...
```

In this case, the customSizing is applied to not only echo container, also sidecar containers (monitoring-daemon).

On the other hand, If we specify the **container** name such as `echo` for customSizing like this:

```yaml
deploymentConfigurations:
  deploymentEnvironment:
    customSizing:
      echo:
        limits:
          cpu: 4
          memory: 3Gi
        requests:
          cpu: 2
          memory: 3Gi
```

Halyard creates Deployment like this:

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: spin-echo
...
spec:
...
  template:
...
    spec:
      containers:
      - name: echo
        resources:
          limits:
            cpu: "4"
            memory: 3Gi
          requests:
            cpu: "2"
            memory: 3Gi
      - name: monitoring-daemon
      # no resources
...
```
the customSizing is applied to only echo container.


Does it make sense?

Thank you!

ezimanyi (approved):
Wow, thanks for the explanation!  I did not know that Halyard made that distinction.  This looks good to me then!

dragon3:
Thank you for the review!

PR #1795, opened 2020-04-13
docs(plugins): Add redirect-from to user and creator guides

aimeeu:
Add redirect-from in front matter to redirect from old page to new page.  Google search for "Spinnaker plugins" has results that point to pages that no longer exist. Added redirect-from so users don't get a 404.

The `jekyll-redirect-from` plugin is listed in `_config.yml` but not listed in the Gemfile. Anybody know why? And will Jekyll load the `jekyll-redirect-plugin` if it's not listed in the Gemfile?

### DIFF
--- a/guides/developer/plugin-creators/overview.md
+++ b/guides/developer/plugin-creators/overview.md
@@ -3,6 +3,8 @@ layout: single
 title: "Overview"
 sidebar:
   nav: guides
+redirect-from:
+  - /guides/developer/plugin-creators/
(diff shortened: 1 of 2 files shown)


317brian:
I don't know the answer to your question @aimeeu, but the syntax in your fix looks correct to me. We'll give it a try and see what happens. The page can't 404 any more than it already is ;)

317brian:
The redirect works

Issue #1546, opened 2019-10-20
CloudFormation stack template does not work


mutabletao:
Page: [/setup/install/providers/aws/aws-ec2/](https://www.spinnaker.io/setup/install/providers/aws/aws-ec2/)

Feedback: running the provided stack template with recommended parameters fails with the message : Unresolved resource dependencies [SpinnakerInstanceProfile] in the Outputs block of the template

dorbin:
@edwinavalos 
@clareliguori 
@anshrma

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

clareliguori:
cc @saimadineni

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

aleon1220:
what i dont like is that these issues never get ressolved. Is there any status on this? I had similar error and the problem was the IAM configuration plus cloud driver was not enabled....

the-mikedavis:
@spinnakerbot remove-label to-be-closed

seeing this with the AWS console & the CLI

```console
$ aws cloudformation deploy --stack-name spinnaker-managing-infrastructure-setup --template-file managing.yaml --parameter-overrides UseAccessKeyForAuthentication=true --capabilities CAPABILITY_NAMED_IAM --region us-east-2
An error occurred (ValidationError) when calling the CreateChangeSet operation: Unresolved resource dependencies [SpinnakerInstanceProfile] in the Outputs block of the template 
```

<img width="1066" alt="Screen Shot 2020-03-10 at 11 45 30 AM" src="https://user-images.githubusercontent.com/21230295/76336931-b485c880-62c4-11ea-9ed6-c03097f5effc.png">

Is there a change that we can make to the `managing.yaml` as a workaround?

roybaryosef-spotinst:
Getting the same issue, couldn't find the proper documentation to work around this.

317brian:
@akshayabd any update on this?

akshayabd:
I've added it in the agenda to discuss in next week's AWS SIG meeting:
[...]

PR #1797, opened 2020-04-15
docs(contributing): Updating AWS EKS install docs

pauljrob:
This is referencing issue: https://github.com/spinnaker/spinnaker.github.io/issues/1546


kaitea:
If we can include details on leveraging Fargate instead of EC2 instances for underlying compute for running the Spinnaker pods on AWS EKS that would be great!
https://aws.amazon.com/blogs/aws/amazon-eks-on-aws-fargate-now-generally-available/

317brian:
Thanks for putting this together @pauljrob! I'll schedule some time to look at this next week.

317brian (approved):

PR #1800, opened 2020-04-16
Removed unnecesary deploy apply step

aleon1220:
removed unnecessary step from **section** Halyard Configurations
@ethanfrogers  Please help me to merge this small fix.

### DIFF
--- a/setup/install/providers/aws/aws-ec2.md
+++ b/setup/install/providers/aws/aws-ec2.md
@@ -97,7 +97,6 @@ After the AWS IAM user, roles, policies and trust relationship have been set up,
 1. Access the Halyard Pod.
 2. Add the configurations for AWS provider with `hal` command. Please check [hal config provider AWS](https://www.spinnaker.io/reference/halyard/commands/#hal-config-provider-aws).
 3. Enable the AWS provider `hal config provider aws enable`.
-4. Apply the configurations to Spinnaker `hal deploy apply`.
 
 ### Configure Halyard to use AccessKeys (if configured)
 


setup/install/providers/aws/aws-ec2.md
```diff
@@ -97,7 +97,6 @@ After the AWS IAM user, roles, policies and trust relationship have been set up,
 1. Access the Halyard Pod.
 2. Add the configurations for AWS provider with `hal` command. Please check [hal config provider AWS](https://www.spinnaker.io/reference/halyard/commands/#hal-config-provider-aws).
 3. Enable the AWS provider `hal config provider aws enable`.
-4. Apply the configurations to Spinnaker `hal deploy apply`.
```
dorbin, line 100: Is this really unnecessary? @ezimanyi 
I though we need to `hal deploy apply` any config changes, such as the `enable` command right above.

ezimanyi: This is necessary; you must do a `hal deploy apply` for any change to the Halyard config to take effect.

dorbin:
I'm closing this for now. Andres, please re-open if I'm missing something.

aleon1220:
Guys i understand your view. Yes it is necessary but not for this point. 
If the steps to install spinnaker are followed from beginning to end then executing `hal deploy apply` after just adding an AWS Provider will throw an error.

Evidence of this is seen in the attached video [Video of a guy following our docs](https://harness.io/2020/02/spinnaker-vs-harness-filming-my-journey/) if you go to the exact time **23:50**

So for this step it should be removed.

ezimanyi:
Oh, I see, if this is intended to be followed from beginning to end then it probably makes sense just to have the user do `hal deploy apply` at the end.  That being said, I don't see any other `hal deploy apply` on the page---would it make sense to add one at the end, or to make it more clear the order of the steps?

Right now it looks like that section is intended to be an overview that tells you the steps to follow (ending in `hal deploy apply`) followed by some more details on the first few steps.  So maybe that's what's confusing, is that the details of those first steps come after the `hal deploy apply`?  Also, as I look more closely at the file, it seems that @akshayabd and @brian-armory added a note to the top that a new version is being worked on---they might actually be in a better place to comment on this.

aleon1220:
> Oh, I see, if this is intended to be followed from beginning to end then it probably makes sense just to have the user do `hal deploy apply` at the end. That being said, I don't see any other `hal deploy apply` on the page---would it make sense to add one at the end, or to make it more clear the order of the steps?
> 
> Right now it looks like that section is intended to be an overview that tells you the steps to follow (ending in `hal deploy apply`) followed by some more details on the first few steps. So maybe that's what's confusing, is that the details of those first steps come after the `hal deploy apply`? Also, as I look more closely at the file, it seems that @akshayabd and @brian-armory added a note to the top that a new version is being worked on---they might actually be in a better place to comment on this.

Correct. I dont know what this new version will contain. I have some ideas on how to improve that page already. 
Maybe we can just clarify that `hal deploy apply` is used only if spinnaker is already installed.

Maybe we can discuss in slack #sig-doc channel @akshayabd and @brian-armory ??

317brian:
Appreciate the discussion here! Generally, I think we should minimize the number of times we tell people to run `hal deploy apply` in a series of steps since there is some wait time after running it.

This PR has the updated AWS docs: https://github.com/spinnaker/spinnaker.github.io/pull/1797. I have not had a chance to review that one yet though.

PR #1802, opened 2020-04-16
add warning regarding invalid Jinja methods

nagonzalez:
There are quite a few examples on how to create dynamic artifacts using `tojson` Jinja method: 
https://github.com/spinnaker/spinnaker/issues/4675
https://github.com/spinnaker/spinnaker/issues/3401

Unfortunately, these snippets no longer work due to Jinjava changes since `1.15`:
https://github.com/spinnaker/orca/pull/2921

We should add warning so folks are aware and can potentially use a workaround:
https://github.com/spinnaker/spinnaker/issues/5012

### DIFF
--- a/reference/artifacts/from-build-triggers.md
+++ b/reference/artifacts/from-build-triggers.md
@@ -87,7 +87,7 @@ customFormat=true
 
 ### Warning
 
-Spinnaker > 1.15.x uses Jinjava 2.2.3 which does not support methods such as `tojson`. Make sure to use supported Jina syntax
+Spinnaker > 1.15.x uses Jinjava 2.2.3 which does not support methods such as `tojson`. Make sure to use supported Jinja syntax
 
 ## Supplied templates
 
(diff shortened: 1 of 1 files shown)


ezimanyi (approved):
I wasn't aware of this change to Jinjava, but seems reasonable to add the warning! Thanks!

nagonzalez:
Yeah, took me quite a while to figure it out. Lemme' know if you folks want to tweak message and/or formatting

reference/artifacts/from-build-triggers.md
```diff
@@ -85,6 +85,10 @@ messageFormat=custom-jar
 customFormat=true
 ```
 
+### Warning
+
+Spinnaker > 1.15.x uses Jinjava 2.2.3 which does not support methods such as `tojson`. Make sure to use supported Jina syntax
```
dbyron0:
```suggestion
Spinnaker > 1.15.x uses Jinjava 2.2.3 which does not support methods such as `tojson`. Make sure to use supported Jinja syntax
```

dorbin (approved):

PR #1807, opened 2020-04-20
feat(docs): document how to hide stages from end-users

maggieneterval:
We [replaced](https://github.com/spinnaker/deck/pull/8175) Deck's stage-specific feature flags with a single `hiddenStages` setting. This change updates instructions for enabling stage-specific flags and adds a new section about how to do this in new versions of Spinnaker.


setup/ci/travis.md
```diff
@@ -25,7 +25,7 @@ repos you should see.
 
    `hal config ci travis enable`
 
-1. Turn on the Travis stage feature:
+1. If using Spinnaker 1.19 or earlier, turn on the Travis stage feature:
```
dorbin:
```suggestion
1. If you're using Spinnaker 1.19 or earlier, turn on the Travis stage feature:
```

setup/ci/wercker.md
```diff
@@ -28,7 +28,7 @@ and credentials.
    hal config ci wercker enable
    ```
 
-2. And that the Wercker stage feature flag is turned on:
+2. If using Spinnaker 1.19 or earlier, make sure that the Wercker stage feature flag is turned on:
```
dorbin:
```suggestion
2. If you're using Spinnaker 1.19 or earlier, make sure that the Wercker stage feature flag is turned on:
```

reference/pipeline/stages/index.md
```diff
@@ -388,3 +388,15 @@ Extend Spinnaker with [predefined Webhook stages](/guides/operator/custom-webhoo
 ### Custom Job
 
 Extend Spinnaker with [predefined Run Job stages](/guides/operator/custom-job-stages/).
+
+## Hiding Stages
```
dorbin: Do you think this might work better under `/guides/operator/` ?

maggieneterval: Makes sense; added it as a standalone page under `/guides/operator`! All comments should be addressed; this is ready for a final look whenever you have a minute 😄

dorbin (approved):
🥇

PR #1813, opened 2020-04-22
feat(plugins): signal that --ui-resource-location will be unneeded in…

danielpeach:
… Spinnaker 1.20.0

These notes are inadequate and will need to be updated once Spinnaker 1.20.0 is actually out, but it's a signal of what's to come. 

In order to omit the `--ui-resource-location` flag from your Hal config, you'll need `pf4jStagePlugin@1.1.3`, but that version requires a change in Orca that won't arrive until Spinnaker 1.20.0.


danielpeach:
@aimeeu please review

aimeeu (approved):
Usually we don't put notes about future releases, but since plugins functionality is still alpha, I'm OK with this.

PR #1814, opened 2020-04-22
feat(docs): consistent naming of the Kubernetes provider

maggieneterval:
Related to: https://github.com/spinnaker/governance/blob/master/rfc/eol_kubernetes_v1.md

The Kubernetes SIG has agreed on standardized naming for Spinnaker's V1 and V2 Kubernetes providers:
- We will refer to the V2 provider as "the Kubernetes provider."
- We will refer to the V1 provider as "the legacy Kubernetes provider."

The primary goal of this standardization is to decrease friction for new users and evaluators, who should not need to understand which sets of terminology ("V2", "Manifest-based") refer to which provider, or even know that there was a legacy provider. A secondary goal is to improve perception of Spinnaker's Kubernetes provider as Kubernetes-native, which was arguably impeded by frequently referring to it as a "Manifest-based" provider, which sounds a bit dated given that applying YAML is the de facto way to deploy to Kubernetes.

### DIFF
--- a/setup/install/providers/kubernetes-v2/index.md
+++ b/setup/install/providers/kubernetes-v2/index.md
@@ -177,11 +177,12 @@ reasons:
   adopt the Kubernetes resources and operations more natively.
   
 * The V2 provider does __not__ use the [Docker Registry
-  Provider](https://www.spinnaker.io/setup/install/providers/docker-registry/), and we
-  encourage you to stop using the Docker Registry accounts in Spinnaker.  The
-  V2 provider requires that you manage your private registry [configuration and
-  authentication](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
-  yourself.
+  Provider](https://www.spinnaker.io/setup/install/providers/docker-registry/).
+  You may still need Docker Registry accounts to trigger pipelines, but
+  otherwise we encourage you to stop using Docker Registry accounts in Spinnaker.
+  The V2 provider requires that you manage your private registry [configuration
+  and authentication](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
+  yourself.  
 
 However, you can easily migrate your _infrastructure_ into the V2 provider.
 For any V1 account you have running, you can add a V2 account following the
--- a/setup/install/providers/kubernetes-v2/index.md
+++ b/setup/install/providers/kubernetes-v2/index.md
@@ -1,6 +1,6 @@
 ---
 layout: single
-title:  "Kubernetes Provider V2 (Manifest Based)"
+title:  "Kubernetes Provider"
 sidebar:
   nav: setup
 ---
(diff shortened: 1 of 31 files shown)


dorbin (approved):
Thanks Maggie! LGTM

ezimanyi (approved):
🚫 ✌️ 2️⃣ 💯 🏅

setup/install/providers/kubernetes-v2/index.md
```diff
@@ -180,6 +175,13 @@ reasons:
   Kubernetes. To avoid building dense and brittle mappings between Spinnaker's
   logical resources and Kubernetes's infrastructure resources, we chose to
   adopt the Kubernetes resources and operations more natively.
+  
+* The V2 provider does __not__ use the [Docker Registry
+  Provider](https://www.spinnaker.io/setup/install/providers/docker-registry/), and we
+  encourage you to stop using the Docker Registry accounts in Spinnaker.  The
```
dbyron0: Don't we still need docker registry accounts to trigger pipelines when new docker images arrive?

maggieneterval: I think the original author's intent here was to clarify that a Kubernetes account no longer requires associated Docker registries (as V1 did to populate Docker images available in the Create Server Group modal), so the Kubernetes provider itself no longer has a dependency on the Docker registry provider...let me know if that makes sense. Suggestions welcome for improved wording, but this diff is just pushing that existing block of text as-is into the bulleted list.

dbyron0: Fair enough.  I'm struggling  as I don't see a place that describes triggering pipelines based on the arrival of docker images.  I don't know my way around this repo very well, and my greps are coming up empty.  I sort of expect to see something in https://github.com/spinnaker/spinnaker.github.io/tree/master/guides/user/pipeline/triggers-with-artifactsrewrite or https://github.com/spinnaker/spinnaker.github.io/tree/master/guides/user/pipeline/triggers.  Maybe it's not the end of the world to have no link for the `trigger pipelines` bit.
```
The V2 provider does __not__ use the [Docker Registry
Provider](https://www.spinnaker.io/setup/install/providers/docker-registry/).
You may still need Docker Registry accounts to trigger pipelines, but
otherwise we encourage you to stop using Docker Registry accounts in Spinnaker.
The V2 provider requires that you manage your private registry [configuration
and
authentication](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
yourself.
```

maggieneterval: Sure, thanks @dbyron0! You're officially a Spinnaker docs writer. 🤓

maggieneterval:
@brian-armory I tagged you as an FYI so that hopefully new Armory docs and blog posts can name the Kubernetes providers consistently with the open source docs. Thank you!

PR #1816, opened 2020-04-24
docs(artifacts): remove references to deleted Halyard flags

plumpy:


reference/artifacts-with-artifactsrewrite/index.md
```diff
@@ -24,10 +24,10 @@ Keep in mind that the artifact in Spinnaker is a _reference_ to an external reso
 
 ## Enabling artifact support
 
-If using a version of Spinnaker prior to 1.20, enable support for the standard artifacts UI:
+If using a version of Spinnaker prior to 1.20, enable support for the standard artifacts UI by editing the `features` section of your Halyard `config` file and adding:
```
maggieneterval, line 27: Oooh sorry, this is confusing because Halyard releases aren't married to Spinnaker releases. Should we qualify this statement by saying "If using a version of Spinnaker prior to 1.20 (and a Halyard version prior to 1.35")?

Or, more in line with what you've suggested here, we can update these instructions to tell users to include `window.spinnakerSettings.feature.artifactsRewrite` in their `settings-local.js`. However, just editing the `features` section of the halconfig won't work post-Halyard-1.35 because we're no longer translating those values into Deck's profile.

dorbin (approved):
LGTM, after Maggie's comment addressed.

reference/artifacts-with-artifactsrewrite/index.md
plumpy: Shooooooot, excellent point. Hm.

What if we revert that Halyard CL for a while and put it back in after 1.21 ships or something? Adding that feature flag to a Spinnaker 1.20+ deployment won't actually cause any issues, right, it'll just get ignored?

Kinda annoying, though, hm...

maggieneterval: Yeah, I'm leaning towards replacing the commands in Halyard until no supported Spinnaker release supports those flags. The main downside, since we don't have versioned Halyard docs, is that the commands will be surfaced for even new users there, leading to the type of feature-flag-overwhelm that was my goal to reduce. But I think in this case it makes sense to not degrade the existing-user experience in favor of reducing confusion in the new-user experience. Going to close this PR and make a separate Halyard PR; will tag you there. Thanks for surfacing this issue 🍡  🤗

PR #1820, opened 2020-04-28
fix(docs): fix Travis and Wercker docs

maggieneterval:
The prior docs would only work with a version of Halyard prior to 1.34. Given that we're going to stop supporting these stage-specific feature flags in 1.20 in favor of the `hiddenStages` setting, which will need to be manually added to Deck's custom profile (not exposed in Halyard), I'd prefer to just update the existing documentation to add the stage-specific flags to Deck's custom profile, as this will work regardless of Halyard version.


ezimanyi (approved):
💯 ⚒️ 📜

PR #1822, opened 2020-04-29
chore(builds): add release badge for 1.20

ethanfrogers:


ttomsu (approved):

ttomsu:
May want to remove 1.17 too

ttomsu (changes requested):
Actually, I have some changes that will make this no longer necessary.

PR #1824, opened 2020-04-29
feat(builds): List releaseN..N-2 statuses, too

ttomsu:
Pulls the branch names into metadata and transforms them as needed. 

The extra dashes in the liquid template `{%- -%}` suppress whitespace, which Markdown tables are sensitive to.

Replaces https://github.com/spinnaker/spinnaker.github.io/pull/1822

Preview:

WuY5gq3aw8e


maggieneterval (approved):
So fancy!

PR #1826, opened 2020-04-29
feat(md): veto docs

emjburns:
Docs on vetoing, plus updating the pinning docs because we don't need type anymore


guides/user/managed-delivery/vetoing/index.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: single
+title: "Vetoing a Version"
+sidebar:
+  nav: guides
+---
+
+{% include toc %}
+
+You can veto a version in an environment to ensure that that version never gets deploy to the environment.
```
erikmunson:
```suggestion
You can veto a version in an environment to ensure that that version never gets deployed to the environment.
```

erikmunson (approved):
short and sweet, small typo fix but no tactical suggestions beyond that.

food for thought though: I think 'mark as bad' resonates with the average user much better than 'veto', which really only makes sense once it's explained to you.

since this is _sort of_ an API doc, I could see a good argument to be made that we should stay true to the API's names and concepts, but if we expect end-users to ever interact with this doc we should think about whether aligning terminology with the UI is good/bad/ugly. I don't have an answer, just posing the question.

emjburns:
Per Erik's feedback I adjusted the wording to be "mark as bad" because we think that might be more intuitive. We might need to change as we get users of this 🤷

PR #1832, opened 2020-05-02
feat: Add API configs section

lerrigatto:
Spinnaker services are configured by yaml files, one each, and usually those files are generated by halyard.
A reference to halyard available configs [exists](https://www.spinnaker.io/reference/halyard/commands/) but it is very hard to match what is generated by it, just looking at the config files themselves.
Finally, halyard doesn't cover every possible available config nor it is really simple to know how to edit each service config manually.

In the context of configuration as code, handling such config files without halyard would be of great help, but in order to do so, better documentation must be put in place.

This PR is a prof of concept of how such documentation can be published, using [kork](https://github.com/spinnaker/kork) as example.
The idea is to fill a folder with the json files generated from the javadocs at build time and display those file in a new section under `/reference/api`.

Given that some services might lack extensive documentation at the moment, this PR should be a good incentive for everybody to increase the amount and quality of comments and documentation. I speak for myself but I would be way more encouraged to write javadocs if they will be accessible to a broad community.

This PR lacks 3 big things:

1. A change in the build process to publish those json files. I envision that the build process could create a PR on this repo with the new set of json and the website can be updated right after. This process should ensure testing and vetting of the new content, either manually or by a bot.
2. The length of this new /reference/api/config section will be huge and we might need to find ways to fraction it. I see 3 options: (i) separate interface like swagger; (ii) huge page with TOC per service like halyard; (iii) each service in a subpage, with a TOC in the index. I personally prefer option (iii)
3. It is unclear to me how many services will be covered from the beginning by this and how standard the generated json file is.

### DIFF
--- a/_data/configs/kork-web.json
+++ b/_data/configs/kork-web.json
@@ -1,0 +1,90 @@
+{
+  "groups": [
+    {
+      "name": "default",
+      "type": "com.netflix.spinnaker.config.TomcatConfigurationProperties",
+      "sourceType": "com.netflix.spinnaker.config.TomcatConfigurationProperties"
+    },
+    {
+      "name": "endpoints.resolved-env",
+      "type": "com.netflix.spinnaker.config.ResolvedEnvironmentConfigurationProperties",
+      "sourceType": "com.netflix.spinnaker.config.ResolvedEnvironmentConfigurationProperties"
+    },
+    {
+      "name": "management.endpoint.resolved-env",
+      "type": "com.netflix.spinnaker.endpoint.ResolvedEnvironmentEndpoint",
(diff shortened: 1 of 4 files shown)


dorbin:
Closing this PR, per discussion in #sig-documentation channel. 

From that discussion:
"You are definitely not alone in wanting better docs around configuration parameters rather than only halyard commands to set a subset of them. As part of the work to create a lighter-weight alternative to Halyard (https://github.com/spinnaker/governance/blob/master/rfc/halyard-lite.md), we are strongly typing each service's configuration parameters in protobuf, from which we can auto-generate documentation. These docs will eventually be auto-published to spinnaker.io; in the meantime, you can check out the work-in-progress docs here: https://github.com/spinnaker/kleat/blob/master/docs/docs.md."

PR #1833, opened 2020-05-04
feat(oauth2): Improve OAuth 2.0 Documentation

ezimanyi:
Currently there is a lot of logic baked into Halyard commands to automatically set fields in the OAuth 2.0 config. While in some cases it might make set-up easier, it comes at the expense of the ability to see and modify what is being configured.

Kleat will not replicate this behavior; it will not store a mapping of the token URI for a number of common OAuth 2.0 providers; users will need to set this themselves in their YAML config. In order to make this easier, let's improve the documentation so that it's easy for users to set up any OAuth 2.0 provider by just pasting in the provided YAML and configuring as necessary.

I think this will actually make it easier for users to troubleshoot as it will be more clear to them what fields they have set (rather than having this done behind the scenes).

This commit:
(1) Adds full documentation on the OAuth 2.0 config; it is currently copy-pasted from the source of truth protos, but once kleat is more mature we can probably just link out to its documentation. (I don't want a consumer of the documentation yet so we are more free to refactor it as we'll likely think about once all the protos are written).
(2) Augments all the inline hal commands with documentation on how to configure something by editing your YAML. (I actually think this will be more clear to operators familiar with YAML than the hal commands.)
(3) Updates the provider-specific documentation to include a block of YAML including what Halyard defaulted and kleat will not default.
(4) Updates references to all say OAuth 2.0 for consistency (fixing the mix of OAuth2 Oauth2 OAuth 2 and OAuth 2.0).

Users using Halyard can use either the YAML-editing or CLI instructions; users using kleat will need to use the YAML-editing instructions. (Though given that kleat is not even alpha yet I have not mentioned it at all here, we'll need to holistically update the docs when it's ready for users.)

### DIFF
--- a/setup/security/authentication/oauth/azure/index.md
+++ b/setup/security/authentication/oauth/azure/index.md
@@ -4,7 +4,9 @@ sidebar:
   nav: setup
 ---
 
-This page instructs you on how to obtain an OAuth 2.0 client ID and client secret for use with your Microsoft Azure tenant.
+This page instructs you on how to obtain an OAuth 2.0 client ID and client secret for
+use with your Microsoft Azure tenant. More extensive documentation is available on
+[Microsoft's site](https://docs.microsoft.com/en-us/azure/active-directory/azuread-dev/v1-protocols-oauth-code).
 
 ## Setting up an Azure Application Registration
 
@@ -19,13 +21,50 @@ This page instructs you on how to obtain an OAuth 2.0 client ID and client secre
 5. Click "Settings" -> "Keys". Under "Passwords", add a Key Description (eg Spinnaker), set the expiry and then click "Save".
    "Value" will now be populated. This is your client-secret, copy it to a safe place.
 
-## Applying the settings in hal
+## Configure Halyard
 
-Set up oauth2 with azure:
+You may configure Halyard either with the CLI or by manually editing the hal config.
 
-`hal config security authn oauth2 edit --provider azure --client-id (client id from above)  --client-secret (client secret from above)`
+### Hal config
 
-The Tenant ID of your organization is required for Azure OAuth2.0 login. To obtain it:
+```yaml
+security:
+  authn:
+    oauth2:
+      enabled: true
+      client:
+        clientId: # client ID from above
+        clientSecret: client-secret # client secret from above
+        accessTokenUri: https://login.microsoftonline.com/${azureTenantId}/oauth2/token
+        userAuthorizationUri: https://login.microsoftonline.com/${azureTenantId}/oauth2/authorize?resource=https://graph.windows.net
+        clientAuthenticationScheme: query
+        scope: profile
+      # You may want to restrict access to your Spinnaker by adding
+      # userInfoRequirements to further restrict access beyond beyond simply
+      # requiring that users have a valid account in your Azure AD Tenant.
+      userInfoRequirements: {}
+      resource:
+        userInfoUri: https://graph.windows.net/me?api-version=1.6
+      userInfoMapping:
+        email: userPrincipalName
+        firstName: givenName
+        lastName: surname
+      provider: AZURE
+```
+
+### CLI
+
+Set up OAuth 2.0 with azure:
+
+`hal config security authn oauth2 edit --provider azure --client-id (client ID from above)  --client-secret (client secret from above)`
+
+Now enable OAuth 2.0 using hal:
+
+`hal config security authn oauth2 enable`
+
+
+## Set environment variable
+The Tenant ID of your organization is required for Azure OAuth 2.0 login. To obtain it:
 1. Navigate to [https://portal.azure.com](https://portal.azure.com) and log in with your Azure credentials.
 2. On the left hand navigation pane, click "Azure Active Directory" --> "Properties".
 3. "Directory ID" is your Tenant ID.
@@ -35,7 +74,3 @@ In order to pass the Tenant ID to gate, we need to insert is as an environment v
 env:
   azureTenantId: (your tenant id)
 ```
-
-Now enable OAuth2 using hal:
-
-`hal config security authn oauth2 enable`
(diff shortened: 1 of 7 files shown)


ezimanyi:
"Don't implement complex technical solutions to problems that can be solved with better documentation."
-Maggie Neterval

"You say a lot of smart things."
-Eric Zimanyi

setup/security/authentication/oauth/config.md
```diff
@@ -0,0 +1,67 @@
...
+      # Whether OAuth 2.0 is enabled.
+      enabled: boolean
+      client:
+        # The OAuth client ID you have configured with your OAuth 2.0 provider.
+        clientId: string
+        
+        # The OAuth client secret you have configured with your OAuth provider.
+        clientSecret: string 
+          
+        # The access token uri for your OAuth provider.
```
maggieneterval:
```suggestion
        # The access token URI for your OAuth provider.
```

setup/security/authentication/oauth/config.md
```diff
@@ -0,0 +1,67 @@
...
+        accessTokenUri: string
+          
+        # The user authorization uri for your OAuth 2.0 provider.
```
maggieneterval:
```suggestion
        # The user authorization URI for your OAuth 2.0 provider.
```

setup/security/authentication/oauth/config.md
maggieneterval: (Just because you also capitalize `URL` and `URI` below)

setup/security/authentication/oauth/google/index.md
```diff
@@ -18,11 +18,41 @@ This page instructs you on how to obtain an OAuth 2.0 client ID and client secre
 ![GCP console to create OAuth 2.0 client screenshot](gcp-oauth-client.png)
 
 
+## Configure Halyard
```
dorbin, line 21: I see this often, "configure Halyard," and I think: you're not configuring Halyard, you're _using_ Halyard to configure Spinnaker. Thoughts?

setup/security/authentication/oauth/index.md
```diff
@@ -11,70 +11,65 @@ by the identity provider. To confirm your identity, Spinnaker requests access to
...
+these fields.
 
-These OAuth 2.0 providers below have been pre-configured in Spinnaker. Follow the instructions to obtain a client ID 
-and client secret.
+## OAuth 2.0 providers
 
-* [Google Apps for Work / G Suite](./google/)
-* [GitHub Teams](./github/)
-* [Azure](./azure/)
+In the general case, you’ll need to consult the documentation of your OAuth 2
```
dorbin:
```suggestion
Consult the documentation of your OAuth 2
```

setup/security/authentication/oauth/index.md
```diff
@@ -11,70 +11,65 @@ by the identity provider. To confirm your identity, Spinnaker requests access to
...
+provider to determine the appropriate values to put in each configurable field. For
+some common OAuth 2.0 providers, specific documentation is provided here.
 
-### Pre-configured providers
-
-For convenience, several providers are already pre-configured. As an administrator, you merely have
- to activate one, and give the client ID and secret. Follow the Provider-Specific documentation to
- obtain your client ID and client secret.
+If you are using one of these providers, please follow the appropriate link
```
dorbin:
```suggestion
If you're using one of these providers, please follow the appropriate link
```

setup/security/authentication/oauth/index.md
```diff
@@ -95,63 +90,28 @@ userInfoMapping:
...
-> Be sure to include the `/login` suffix at the end of the `--pre-established-redirect-uri` flag!
+## Restricting access based on User Info
 
-Additionally, some configurations make it necessary to "unwind" external proxy instances. This makes the request to Gate
-look like the original request to the outer-most proxy. Add this to your `gate-local.yml` file in your Halyard
-[custom profile](/reference/halyard/custom/#custom-profiles):
+User access can be restricted further based on the user info from an OAuth ID token. This
+requirement is set via the `security.authn.oauth2.userInfoRequirements` field, which
+is a map of key/value pairs. The values are interpreted as regular expressions if
+if they start and end with '/'. This enables us to restrict user login to specific
```
dorbin:
```suggestion
they start and end with '/'. This enables us to restrict user login to specific
```

dorbin: "us"?

dorbin (approved):
Just a few minor things, but LGTM

setup/security/authentication/oauth/azure/index.md
```diff
@@ -19,13 +21,50 @@ This page instructs you on how to obtain an OAuth 2.0 client ID and client secre
...
 
-The Tenant ID of your organization is required for Azure OAuth2.0 login. To obtain it:
+```yaml
+security:
+  authn:
+    oauth2:
+      enabled: true
+      client:
+        clientId: # client ID from above
+        clientSecret: client-secret # client secret from above
```
maggieneterval: This is super minor, but the first thing I did when I read "client ID from above" is ctrl+f for "client ID" on this page, and initially couldn't find it because in step 4 it says "this is the client-id to pass to hal." It won't let me comment on that line since it's not one you edited, but maybe we can change client-id to client ID so it matches what's here (same with client-secret --> client secret?).

maggieneterval (approved):
📈 ⭕ 🔐 📝  💯 
🚫 🔢 ✅ 📝

setup/security/authentication/oauth/google/index.md
ezimanyi: I suppose, though to the extent that Halyard is an abstraction over configuring Spinnaker directly, I still think of it as configuring Halyard (then using Halyard to *deploy* Spinnaker).  But I'm also happy to use a different term here...I just can't really think of one that's better.

setup/security/authentication/oauth/index.md
ezimanyi: I'd moved this from below, so that's pre-existing, but agree it's unclear who "us" is; changed to use the gerund form which eliminates an ambiguous subject.

setup/security/authentication/oauth/azure/index.md
ezimanyi: Good catch! Fixed!

setup/security/authentication/oauth/google/index.md
dorbin: Aight, stet. It's a small thing anyway. :)

PR #1838, opened 2020-05-06
feat(docs): remove skin documentation

maggieneterval:
With the removal of the Kubernetes V1 provider, no providers implement a non-default `skin`, and Deck will no longer expose `skin` as an extension point in 1.21. This should not be hugely controversial, since to the best of our knowledge, the Kubernetes providers were the only client of the skin API. Let's remove the documentation for creating custom CRD UIs using skins, especially given that the Kubernetes SIG is discussing a more holistic solution to exposing information about CRDs that hopefully does not require a custom build.

Corresponding changes:
- [Deck](https://github.com/spinnaker/deck/pull/8263)
- [Gate](https://github.com/spinnaker/gate/pull/1186)
- [Clouddriver](https://github.com/spinnaker/clouddriver/pull/4576)


ezimanyi (approved):
💯 🥳 
Thank you again for thinking about keeping the docs up to date (in addition to the great clean-up work in the code)!

dorbin (approved):

ethanfrogers (approved):

PR #1843, opened 2020-05-09
docs(orca): Moved Maria DB set up after Orca SQL title

aleon1220:
Moved MariaDB set up before Netflix example
it makes more sense to be placed there
Example is at the end
Added a line to separate the instructions from the example


317brian (approved):

aleon1220:
thanks

PR #1850, opened 2020-05-14
fix(docs): fix broken links in crd-extensions

glauco:


dorbin (approved):
Thanks for finding and fixing these, @glauco

PR #1853, opened 2020-05-19
Added the set command and reference to mysql docs

aleon1220:
added link to mysql docs
added set command to make it explicit where it has to run

### DIFF
--- a/setup/productionize/persistence/orca-sql.md
+++ b/setup/productionize/persistence/orca-sql.md
@@ -27,7 +27,7 @@ The `migration` user will only be used to perform schema changes on the database
 
 Before deploying Orca, the schema and database uses must first be manually setup:
 
-1. Set MySQL Server variable `tx_isolation` setting to `READ-COMMITTED`. Refer to [MySQL Server System Variables](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation)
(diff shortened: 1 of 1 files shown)


setup/productionize/persistence/orca-sql.md
```diff
@@ -27,7 +27,13 @@ The `migration` user will only be used to perform schema changes on the database
 
 Before deploying Orca, the schema and database uses must first be manually setup:
 
-1. Set MySQL Server `tx_isolation` setting to `READ-COMMITTED`
+1. Set MySQL Server variable `tx_isolation` setting to `READ-COMMITTED`. Refer to [MySQL Server System Variables](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation)
```
dorbin: This page times out for me. Also, there's a version number in the url. Will the link become obsolete?

dorbin:
```suggestion
1. Set MySQL Server variable `tx_isolation` setting to `READ-COMMITTED`. Refer to [MySQL Server System Variables](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation).
```

aleon1220: yes is for version 5.7 only

dorbin: Thanks, I hadn't noticed the heading above, _supposed to be_ for 5.7 only. :)

And the url now works for me.

dorbin (approved):

PR #1854, opened 2020-05-19
Docs(setup): adds system hardware requirements to setup overview

### LINKED ISSUE #1852: Hardware requirements for Spinnaker?
Page: [/setup/install/halyard/](https://www.spinnaker.io/setup/install/halyard/)

Feedback:

I have gone through the setup instruction, its very hard to find the hardware requirements.
Can you document the requirements in the Setup section?

dorbin:
Closes #1852

### DIFF
--- a/setup/install/index.md
+++ b/setup/install/index.md
@@ -15,9 +15,13 @@ in [Quickstart](/setup/quickstart/) might be a better choice.
 * A machine on which to install Halyard
 
   This can be a local machine or VM (Ubuntu 14.04/16.04, Debian, or macOS), or
-  it can be a Docker container.
+  it can be a Docker container. Make sure it has at least 4GB of memory.
 
 * A Kubernetes cluster on which to install Spinnaker itself
+
+  We recommend at least 4 cores and 8GB of RAM available in the cluster. 
+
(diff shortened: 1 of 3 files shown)


setup/install/index.md
```diff
@@ -15,10 +15,14 @@ in [Quickstart](/setup/quickstart/) might be a better choice.
 * A machine on which to install Halyard
 
   This can be a local machine or VM (Ubuntu 14.04/16.04, Debian, or macOS), or
-  it can be a Docker container.
+  it can be a Docker container. Make sure it has at least 4GB of memory.
 
 * A Kubernetes cluster on which to install Spinnaker itself
 
+  We recommend at least 4 cores and 8GB of RAM available in the cluster.
```
ezimanyi: I would suggest at least 16 GB of memory for the machine (or cluster) that will run Spinnaker.  I've managed to get it running on a machine with 8 GB but memory was very low, and led to some instability, so I would not recommend that as a starting point.  Maybe 12 GB is enough if we don't want to over-provision, but I think people will be frustrated if we tell them to provision 8 GB.

ezimanyi (approved):

PR #1856, opened 2020-05-27
feat(docs): add legacy artifacts UI migration docs

maggieneterval:
Addresses: https://github.com/spinnaker/spinnaker/issues/5761

Let's add a specific section for what to expect when upgrading to 1.20 after using the legacy artifacts UI, specifically calling out the only reported pain point so far, which is the absence of a separate Expected Artifacts section.


ezimanyi (approved):
🏺 📜 💯

reference/artifacts-with-artifactsrewrite/index.md
```diff
@@ -32,6 +32,23 @@ hal config features edit --artifacts-rewrite true
 
 If using Spinnaker 1.20 or later, support for the standard artifacts UI is enabled by default.
 
+## Migrating from the legacy artifacts UI
```
dorbin: @maggieneterval is "migrating" the right word here? This sounds like you're saying...
* here's what's changed and how to deal with it
* here's how to re-enable legacy

maggieneterval: Fair point, how do you feel about "Transitioning from the legacy artifacts UI" instead?

dorbin: :100:

dorbin (approved):

PR #1859, opened 2020-05-28
docs(canary): Fix typo.

smstone:


dorbin (approved):
good catch, @smstone, Thanks!

PR #1860, opened 2020-05-28
feat(docs): remove obsolete codelab

maggieneterval:
This codelab is so old that it doesn't even describe the legacy Kubernetes provider as V1. I only found this because it was referenced in our onboarding docs.


ezimanyi (approved):
🚫 🔡 👩‍🔬 💯 ✅ 
🔛 🎓 📈

dorbin (approved):
Thanks, Maggie. This is great!

PR #1874, opened 2020-06-11
Dnilasor home garden banner update

dnilasor:
I'd like to update the spinnaker gardening days advertisement on the home page. This version contains the new registration link for the July event, along with updated copy and new event branding.

Re: @dorbin inquiry, updates to the community/gardening are to reflect the upcoming event are forthcoming, including better long-term organization (will also re: your msg)


dorbin (approved):
🥇

PR #1640, opened 2019-12-17
Clarify that the 1.18 release will be used to fine-tune stats collection

demobox:

PR #1878, opened 2020-06-16
fix(core): Update some terminology

ezimanyi:
Update the few places where we use the term 'whitelist' to better terminology.

### DIFF
--- a/community/stats/index.md
+++ b/community/stats/index.md
@@ -33,7 +33,7 @@ Spinnaker's telemetry module collects the following bits of data:
   * Each stage's:
     * Final status
     * Type
-    * Cloud Provider (if whitelisted)
+    * Cloud Provider
 
 Spinnaker's telemetry project has been reviewed and approved as compliant with the [Linux Foundation's Telemetry Data Collection and Usage Policy](https://www.linuxfoundation.org/telemetry-data-policy/#completed).
 
--- a/reference/pipeline/expressions/index.md
+++ b/reference/pipeline/expressions/index.md
@@ -26,8 +26,8 @@ be evaluated.
 Spinnaker allows you to execute Java code within a pipeline expression. This can
 be useful for string manipulation or more advanced custom logic than
 would otherwise be possible. For security reasons, you can only call methods of
-whitelisted Java classes. You can find the full list of whitelisted classes
-[here](#whitelisted-java-classes).
+specifically allowed Java classes. You can find the full list of allowed classes
+[here](#allowed-java-classes).
 
 You can use methods directly on existing map values based on their Java class.
 For example, you can process a list of values entered as a parameter using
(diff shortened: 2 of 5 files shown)


community/stats/index.md
```diff
@@ -33,7 +33,7 @@ Spinnaker's telemetry module collects the following bits of data:
   * Each stage's:
     * Final status
     * Type
-    * Cloud Provider (if whitelisted)
+    * Cloud Provider
```
ezimanyi, line 36: I just removed this particular instance, as it didn't seem like it was providing useful/helpful information.

In all of these cases (status, type, cloud provider) we only send along the field if we recognize it (ie, if it's a cloud provider we've heard of or a stage type we know about) and send "UNKNOWN" otherwise.  Maybe there's a way of saying that in a separate paragraph that makes it clear, but as it was I didn't really understand what that meant without looking at the code (and it applies to more than just the cloud provider).

reference/pipeline/expressions/index.md
```diff
@@ -261,24 +261,27 @@ For example, `${#triggerResolvedArtifact("my-image")["reference"]}` might return
 A shortcut to look up the resolved artifact in execution trigger by its type. If multiple artifacts are found, only 1 will be returned.
 For example, `${#triggerResolvedArtifactByType("docker/image")["reference"]}` might return `gcr.io/spinnaker-marketplace/orca@sha256:b48dbe7d7cb580db8512e4687d31f3710185b08afcf3cb53c0203025f93f9091`.
 
-## Whitelisted Java classes
+## Allowed Java classes
 
-You can find the code which whitelists Java classes [here](https://github.com/spinnaker/orca/blob/6d0ba0bf8af5e06c5b405b8294f07e7a5a4c335a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/whitelisting/InstantiationTypeRestrictor.java#L26).
-The whitelisted classes are:
+You can find the code which restricts instantiable Java classes [here](https://github.com/spinnaker/kork/blob/8e414e21625219a58afece0a812f570340882294/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/whitelisting/InstantiationTypeRestrictor.java#L28).
+The allowed classes are:
```
ezimanyi, line 267: I updated the link to point to where the code currently lives, and also added three more classes that have been added to the code.

setup/monitoring/stackdriver.md
```diff
@@ -108,11 +108,6 @@ Proceed to [install the dashboards](#installing-the-stackdriver-dashboards)
 
 ## Installing the Stackdriver dashboards
 
-*__Note__: The Stackdriver Dashboard API is currently whitelisted, so
-you need a registered STACKDRIVER_API key in order to use it. If you
-are so inclined, [contact your sales
-representative](https://cloud.google.com/contact/){:target="\_blank"}.
-
```
ezimanyi, line 115: This caveat is 3 years old and appears to no longer be true; as far as I can tell this is [GA now](https://cloud.google.com/monitoring/dashboards/api-dashboard).

setup/security/authorization/index.md
```diff
@@ -29,7 +29,7 @@ Permissions can be attached to applications and (provider) accounts. A permissio
 
 Keep these in mind as you consider your authorization strategy:
 
-1) Fiat's authorization model is a whitelist that is open by default. In other words, when a resource does _not_ 
+1) Fiat's authorization model open by default. In other words, when a resource does _not_
```
ezimanyi, line 32: In this case, I did not replace with another term as it seems like the point of this paragraph is that it is open by default, and didn't want to detract from that.

markjacksonfishing (approved):
non-binding +1

community/stats/index.md
dorbin: Good point that this applies to all these items, not just provider. 
It [_looks like_](https://github.com/spinnaker/spinnaker.github.io/pull/1640) this is a setting that can be enabled/disabled (enabled by default, as of 1.19).

dorbin (approved):
I just had one question about the "Cloud Provider (if whitelisted)" part, but am approving.

I'm gonna add to the Doc SIG agenda that we consider addressing whitelist/blacklist terminolgy sitewide.

community/stats/index.md
ezimanyi: Yes, stats collection is indeed controllable by a setting (which is now enabled by default).

The issue here is whether, if someone built a custom cloud provider named `my-sensitive-cloud-provider`, if we would send that cloud provider name as part of the stats collection.   (Similarly if I made a custom stage called `eric-custom-stage-secret`, would we send that along.)

The answer in both cases is that we won't, because we have a list of known providers/stages and the client code only sends along those we know about, and just sends along `UNKNOWN` if it's not one in that list.

Let me know if you think we should add something to this effect in that section; I'm not sure if this is helpful information or just going to confuse people.

ezimanyi:
This was all places I could find in the docs that use the terminology, so unless I missed something, there shouldn't be any other follow-up work needed.

ezimanyi:
Will merge this, but happy to circle back and add a clarification on the stats item.

PR #1883, opened 2020-06-18
chore(docs): Fix misnamed property listed in orca sql configuration

clanesf:
While going through some recent database troubleshooting, it came to my attention that the property `connectionTimeout` on this page should actually be `connectionTimeoutMs`.

Using just `connectionTimeout` does not work which is only noticable if you are trying to increase the timeout because you are hitting it and realize that adjusting this value doesn't do anything. Correcting this in case anyone else runs into similar scaling issues and needs to be able to adjust this value.

Property names are taken from https://github.com/spinnaker/kork/blob/master/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/ConnectionPoolProperties.kt

It's also correct on the page https://github.com/spinnaker/orca/blob/7c4d3ceece4063dc62e46913b62da67dfddd452b/orca-peering/README.md#operating-notes but somehow it was overlooked here.


setup/productionize/persistence/orca-sql.md
```diff
@@ -64,7 +64,8 @@ sql:
     jdbcUrl: jdbc:mysql://localhost:3306/orca
     user: orca_service
     password: hunter2
-    connectionTimeout: 5000
+    connectionTimeoutMs: 5000 
+    maxLifetimeMs: 30000
     maxLifetime: 30000
```
robzienert, line 69: Looks like you're missing the removal of `maxLifetime`.

robzienert (changes requested):

robzienert:
Looks great after the comment is fixed. Thanks!

PR #1885, opened 2020-06-18
fix(plugins): remove edit plugin command docs

edgarulg:
I am removing the documentation of edit plugin command because that functionality no longer exist on Halyard CLI. Here is the PR for remove it: https://github.com/spinnaker/halyard/pull/1678


link108 (approved):
Thank you @edgarulg!

PR #1898, opened 2020-07-02
docs(kubernetes): Add release notes about job suffix change

ezimanyi:
The 1.22 release will change the default for whether we add a suffix to Kubernetes jobs. Add a release note about this.

### DIFF
--- a/community/releases/next-release-preview/index.md
+++ b/community/releases/next-release-preview/index.md
@@ -12,3 +12,38 @@ in the next release of Spinnaker. These notes will be prepended to the release
 changelog.
 
 ## Coming Soon in Release 1.22
+
+### (Breaking Change) Suffix no longer added to jobs created by Kubernetes _Run Job_ stage
+
+Spinnaker no longer automatically appends a unique suffix to the name of jobs
+created by the Kubernetes _Run Job_ stage. Prior to this release, if you
+specified `metadata.name: my-job`, Spinnaker would update the name to
+`my-job-[random-string]` before deploying the job to Kubernetes. As of this
+release, the job's name will be passed through to Kubernetes exactly as
+supplied.
+
+To continue having a random suffix added to the job name, set the
+`metadata.generateName` field instead of `metadata.name`, which causes the
+[Kubernetes API](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values)
+to append a random suffix to the name.
+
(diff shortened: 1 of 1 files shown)


ezimanyi:
I also realize that I didn't add a note about this in the 1.21 release notes.  While there was no explicit change about this in the 1.21 release notes, I think it makes sense to add something for visibility.  Barring objections, I'll edit the gist for the 1.21 release and add the following.  (While a few people may have already read the notes and will miss this, we're early enough in the cycle that it's probably worth advertising this more.)

### (Upcoming Breaking Change) Suffix no longer added to jobs created by Kubernetes _Run Job_ stage

In 1.22, Spinnaker will no longer automatically append a unique suffix to the name of jobs created by the Kubernetes _Run Job_ stage. Currently, if you specify `metadata.name: my-job`, Spinnaker updates the name to `my-job-[random-string]` before deploying the job to Kubernetes. As of Spinnaker 1.22, the job's name will be passed through to Kubernetes exactly as supplied.

To continue having a random suffix added to the job name, users should specify the `metadata.generateName` field instead of `metadata.name`, which will cause the [Kubernetes API](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values) to append a random suffix to the name.

This change is particularly important for users who are using the preconfigured job stage for Kubernetes, or who are otherwise sharing job stages among different pipelines. In these cases, jobs are often running concurrently, and it is important that each job have a unique name. In order to retain the current behavior, these users will need to manually update their Kubernetes job manifests to use the `generateName` field.

Users of Spinnaker 1.21 can opt in to this new behavior by setting `kubernetes.jobs.append-suffix: false` in their `clouddriver-local.yml`.  This behavior will become the default in Spinnaker 1.22 with an ability to opt out.  The ability to opt out will be removed in Spinnaker 1.23.

Users are encouraged to update any necessary jobs to specify `metadata.generateName` between now and Spinnaker 1.22.  Specifying a `metadata.generateName` will cause a suffix to be appended, regardless of the above flag, so users can incrementally update their jobs and then opt in to the new behavior once all jobs have been updated.  It is *not* necessary to set the above flag before starting to use the `metadata.generateName` field.

maggieneterval (approved):
📝 🏃 🔨 📛 ➕ 🔤 ✅ 

Sounds good re: editing the 1.21 curated changelog as well!

PR #1905, opened 2020-07-08
minor typo 'nodeSelector' --> 'nodeSelectors'

shubhamc183:
`nodeSelectors` is expected in hal config instead of `nodeSelector`

image

Results into error

image

`nodeSelectors` works fine but not actually respected.


plumpy (approved):
thanks so much for fixing this!

PR #1929, opened 2020-07-17
docs(fix): add content to section index pages

### LINKED ISSUE #1925: Plugin creation page is empty
Page: [/guides/developer/plugin-creators/](https://www.spinnaker.io/guides/developer/plugin-creators/)

Feedback: Plugin creation guide page is empty. Step by step guide should be created (screenshots and code snippets would be nice)

aimeeu:
Add content to section index pages in case users click on the breadcrumbs instead of the left navigation tree.

Fixes: https://github.com/spinnaker/spinnaker.github.io/issues/1925

### DIFF
--- a/guides/developer/dev-env/index.md
+++ b/guides/developer/dev-env/index.md
@@ -5,3 +5,9 @@ sidebar:
   nav: guides
 ---
 
+This section contains the following content:
+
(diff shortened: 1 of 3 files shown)


dorbin (approved):
Nit: the order of the subsections doesn't match the order they're listed in the nav tree. I don't think that's a very big deal.

PR #1930, opened 2020-07-17
docs(fix): Add redirect to fix plugins/user-guide 404

aimeeu:
Google search turns up this result:   https://spinnaker.io/guides/user/plugins/user-guide/  which gives a 404


dorbin (approved):
Thanks, Aimee!

PR #1931, opened 2020-07-17
docs(gardening): Removing -n namespace

EasyAsABC123:
It is unneeded because it is defined in the garden.yaml provided by armory


aimeeu (approved):

aimeeu:
Thanks @EasyAsABC123

PR #1938, opened 2020-07-20
Change serviceAccount role to use gcloud bucket

abdulbasitkay:
Change role to use gcloud bucket with less privilege but enough for the task. this is as suggested in  [#5235](https://github.com/spinnaker/spinnaker/issues/5235)


maggieneterval (commented):
Hi @abdulbasitkay, thanks so much for the PR!

Only granting the service account the least privileged role necessary makes sense to me. Out of curiosity, do you think an even less-privileged role like `objectViewer` would work, or does Spinnaker's Docker provider require more than read access?

abdulbasitkay:
That's something I wasn't sure of. I would have to reach out to someone with more context better understand it.
Or @maggieneterval do you have some context to advise?

maggieneterval:
> That's something I wasn't sure of. I would have to reach out to someone with more context better understand it.
Or @maggieneterval do you have some context to advise?

My guess would be that read-only privileges would be sufficient, since my understanding is that Spinnaker's Docker provider is limited to reading available image tags, and triggering pipelines on image push. We would definitely want to test both populating images and triggering using the less-privileged SA just to be safe, before merging this: is this something you would be able to do? Thanks!

abdulbasitkay:
> 
> 
> > That's something I wasn't sure of. I would have to reach out to someone with more context better understand it.
> > Or @maggieneterval do you have some context to advise?
> 
> My guess would be that read-only privileges would be sufficient, since my understanding is that Spinnaker's Docker provider is limited to reading available image tags, and triggering pipelines on image push. We would definitely want to test both populating images and triggering using the less-privileged SA just to be safe, before merging this: is this something you would be able to do? Thanks!

I've updated the PR. And I've also I've gotten a similar response on what the provider does. 
Unfortunately, don't have everything (GCP) setup to be able to test this.

maggieneterval:
> Unfortunately, don't have everything (GCP) setup to be able to test this.

Okay, well we definitely want to confirm that the updated setup instructions work before we publish them on our docs site; were you planning on setting up a test Spinnaker with GCP credentials in the near future? If not we'll have to close this until someone is able to test it; I'm happy to do so next time I'm reconfiguring my dev Spinnaker but have several other priorities this week.

abdulbasitkay:
I already have spinnaker setup I guess what would be next is to setup GCP, I'll see how I can do that within the remaining time of the gardening days.

dnilasor (approved):
I do not have firsthand knowledge of this endpoint but this accurately reflects what is proposed in the issue

AugustasV:
@abdulbasitkay Our team have spinnaker cluster on GCP. I could help to test this issue.

devorbitus:
@maggieneterval I have experience with this configuration and I can confirm Spinnaker only needs object viewer permissions to be able to trigger a pipeline when a new docker image is added to the registry. I agree that the proposed change to objectViewer permissions makes the most sense as that is all that Spinnaker really needs to view the contents of the Google Container Registry. 

Technically, we could dive even further by placing the permissions directly on the Google Cloud Storage bucket that is created when you enable GCR but that bucket or buckets will be different based on what global region you are wanting your GCR to reside but I think that complexity should be left to the ultra-security conscious organization rather than people wanting to run simple tutorials.

maggieneterval:
Awesome, thanks so much for confirming, @devorbitus! 🎉

PR #1948, opened 2020-07-21
Fix Fiat's "Administrator functionality" sample profile 

KeisukeYamashita:
## What

As title. If we `hal deploy apply` with the sample profile after placing the file in `~/.hal/$DEPLOYMENT/profiles/fiat-local.yml` as the note suggests, it will not work.


KeisukeYamashita:
Sorry, this was problem with my caching.. the current document is correct. Closing...

PR #1963, opened 2020-07-22
Kustomize document update due to kustomize upgrade

abdulbasitkay:
docs(kustomize): update due to kustomize upgrade

The version of kustomize was upgraded which meant that some information on this docs was no longer accurate.
The version upgrade is done in PR [rosco#685](https://github.com/spinnaker/rosco/pull/685) so this PR, if accepted, should be merged after that.


dnilasor (approved):
The other PR has been merged, and this LGTM

PR #1966, opened 2020-07-22
docs(plugins): Add note - no auth token support

aimeeu:
This quick fix is a result of a [Slack conversation](https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1595322822321400). 

"Token auth isn't supported right now. You might want to try hosting your repository.json and plugins.json file somewhere else in the meantime, S3 perhaps?"


plumpy:
I don't know anything about plugins, assigning this to Cameron who wrote the slack message in question to double-check this :grin: 

Thanks, Aimee!

link108:
LGTM! I think we'd like to support that at some point, but we don't atm

link108 (approved):

PR #1967, opened 2020-07-22
Update index.md

stephly07:
Suggestion to simplify a sentence to make it shorter.
Another suggestion I have is to add additional content to the sentence "It's a good idea though to communicate to the team what templates are available." but I don't exactly know the real reason why it is good.

Is it because it will save time for others having to recreate something that may already exist? Sharing why it's good may encourage people to communicate more.

### DIFF
--- a/guides/user/pipeline/pipeline-templates/index.md
+++ b/guides/user/pipeline/pipeline-templates/index.md
@@ -34,9 +34,9 @@ on an existing pipeline.
 
 * Share the template with one or more teams of developers using Spinnaker.
 
-  To [save a pipeline
+  [Saving a pipeline
   template](/guides/user/pipeline/pipeline-templates/create/#4-save-the-template)
-  to Spinnaker is to make it available to developers. It's a good idea though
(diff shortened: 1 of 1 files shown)


guides/user/pipeline/pipeline-templates/index.md
```diff
@@ -34,9 +34,9 @@ on an existing pipeline.
 
 * Share the template with one or more teams of developers using Spinnaker.
 
-  To [save a pipeline
+  [Saving a pipeline
```
317brian, line 37: The rest of the list starts w/ a verb, such as use. Can you rewrite the sentence to be start w/ "Save" so that it's consistent with the others?

stephly07: Hi Brian,

The rest of the list does start with a verb, but I thought this sentence in particular is more of a note to provide additional context to the second bullet. If this sentence should be an action / step, then I think this should appear as the bullet instead. 

Ex.

- Save a pipeline template to Spinnaker.

This will make the template available to developers. It's a good idea to communicate to the team about what templates are available.

PR #1968, opened 2020-07-22
Update instantiate.md

stephly07:
Bolded the word Note and capitalized the first word in that note
Adjusted the punctuation marks in a sentence to reduce the number of commas
Identified a sentence that may have a missing link

### DIFF
--- a/guides/user/pipeline/pipeline-templates/instantiate.md
+++ b/guides/user/pipeline/pipeline-templates/instantiate.md
@@ -74,12 +74,12 @@ and get it using the following command:
 
 ## Provide values for the variables
 
-> Note: the variables [defined in the pipeline
+> __Note__: The variables [defined in the pipeline
 > template](/guides/user/pipeline/pipeline-templates/create/#3-edit-the-file-for-template-format)
 > include default values, so you don't have to provide a value for every variable defined.
 
-In the pipeline JSON file, in the `variables` section, list each variable
-for which you're providing values, and write that value.
+In the pipeline JSON file (in the `variables` section), list each variable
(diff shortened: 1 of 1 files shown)


guides/user/pipeline/pipeline-templates/instantiate.md
```diff
@@ -91,7 +91,7 @@ for which you're providing values, and write that value.
    ```
 You can code each value by hand in the pipeline JSON that you create. You can
 also generate the JSON and populate the values programatically. For simplicity
-This doc describes doing it by hand.
+This doc describes doing it by hand. <!-- This sentence sounds like "This doc" should be hyperlinked but I don't see anything. I wanted to bring attention to this in case there is supposed to be a link -->
```
dorbin: It's referring to the doc that it's in--it's self referential! But I agree it's a little ambiguous. I thing if it were at the beginning of the article, it would be clear. I'll suggest an edit.

dorbin:
```suggestion
This doc describes doing it by hand. 
```

dorbin: Actually, @stephly07 , in the interest of getting as many of these PRs merged as we can, and because it's not too terribly ambiguous, I'm going to let it ride for now.

dorbin (approved):

guides/user/pipeline/pipeline-templates/instantiate.md
stephly07: Sounds good!

PR #1983, opened 2020-07-23
from Distributed installation to Kubernetes

aleon1220:
I changed to match the convention the other titles follow
$Installation_Type $System_Type
e.g.
Distributed Kubernetes
Local Debian
Local Git

### DIFF
--- a/setup/install/environment.md
+++ b/setup/install/environment.md
@@ -12,7 +12,7 @@ In this step, you tell Halyard in what type of environment to install Spinnaker.
 The recommended path is a distributed installation onto a Kubernetes cluster,
 but all of these methods are supported:
 
-* [Distributed installation](#distributed-installation) on Kubernetes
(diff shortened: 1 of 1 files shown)


setup/install/environment.md
```diff
@@ -12,7 +12,7 @@ In this step, you tell Halyard in what type of environment to install Spinnaker.
 The recommended path is a distributed installation onto a Kubernetes cluster,
 but all of these methods are supported:
 
-* [Distributed installation](#distributed-installation) on Kubernetes
+* [Distributed Kubernetes](#distributed-kubernetes) on Kubernetes
```
dorbin: This should work, but it failed the rake test. Can you take a look, Andres?

aleon1220: i think it might be related to the change in the title. Something in the navegation broke. Is a very small change. I will not change the link only the tile lets see.

aleon1220 (commented):
The change in the link broke the build.

PR #1990, opened 2020-07-29
docs(kubernetes): Update kubernetes v2 provider documentation to state that Spinnaker by default versions ReplicaSets. 

mattsanta:
docs(kubernetes): Update kubernetes v2 provider documentation to state that Spinnaker by default versions ReplicaSets.


mattsanta:
I checked all the kubernetes resource handlers in Clouddriver and ReplicaSet was the only one that's default versioning was wrong in the documentation.

[KubernetesReplicaSetHandler versioning](https://github.com/spinnaker/clouddriver/blob/36235837d452f251a4f50c6bf3d876e6eabdec9c/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesReplicaSetHandler.java#L83)

@maggieneterval FYI since we talked about this :)

maggieneterval (approved):
💯  Thank you for auditing and updating these docs!

PR #2001, opened 2020-08-14
docs(clouddriver/sql): Add release notes for fix to clouddriver sql cache. 

mattsanta:
docs(clouddriver/sql): Add release notes for fix to clouddriver sql cache. 

Release notes for Clouddriver SQL Cache data too long issue fix (https://github.com/spinnaker/spinnaker/issues/5600). Since we've moved to a new version of the tables used for caching we need to instruct users how to delete the old version once they are ready.

PR #2004, opened 2020-08-18
docs(clouddriver/sql): Reverting release note info for Clouddriver SQL cache issue since the fix had to be reverted and won't be released in 1.22. 

mattsanta:
* docs(clouddriver/sql): Reverting release note info for Clouddriver SQL cache issue since the fix had to be reverted and won't be released in 1.22. 

  Revert "docs(clouddriver/sql): Add release notes for fix to clouddriver sql cache.  (#2001)"

  This reverts commit 4a743465228f121db582e3e45a7a19bb2a7d71ef.


mattsanta:
@maggieneterval, also reverting doc change since the fix won't make it in by 1.22.

maggieneterval (approved):

PR #2014, opened 2020-08-26
docs(plugins): add workaround for all services restarting

aimeeu:
When Halyard adds a plugin to a Spinnaker installation, it adds the plugin repository information to each service so the service can download the plugin and check if an extension exists for that service. This causes each service to restart, which is not good for large spinnaker installations.

Halyard workaround -> define plugins in profiles directory for each service (i.e. <service>-local.yml)

Also fixed spacing in deck-proxy YAML snippets


guides/user/plugins/index.md
```diff
@@ -123,6 +123,8 @@ See the command [reference](/reference/halyard/commands/#hal-plugins-repository)
 
 ## Add a plugin using Halyard
 
+>Note: When Halyard adds a plugin to a Spinnaker installation, it adds the plugin repository information to each service. This means each service  downloads the plugin and checks if an extension exists for that service. Each service restarts, which is not good for large Spinnaker installations. See the [Plugin configuration without Halyard](#plugin-configuration-without-halyard) section for how to avoid each service restarting.
```
dorbin:
```suggestion
>Note: When Halyard adds a plugin to a Spinnaker installation, it adds the plugin repository information to each service. This means each service downloads the plugin and checks if an extension exists for that service. Each service restarts, which is not good for large Spinnaker installations. See [Plugin configuration without Halyard](#plugin-configuration-without-halyard) for how to avoid each service restarting.
```

dorbin: "...which is not good for large..." is a value judgment, and probably an accurate one, but can we describe the specific consquences?

guides/user/plugins/index.md
```diff
@@ -200,6 +202,33 @@ spinnaker:
...
+        extensions:
+          <extension-name>:
+            id: <extension-name>
+            enabled: <true-false>
+            config: {}
+```
+
+The plugin developer should provide configuration details in YAML format. If not:
+
+1. Add the plugin using Halyard
```
dorbin: Periods at the ends of each of these

guides/user/plugins/index.md
```diff
@@ -200,6 +202,33 @@ spinnaker:
 
 Note: `hal plugins enable` and `hal plugins disable` enable or disable _all_ plugins, so use with caution.
 
+### Plugin configuration without Halyard
+
+To avoid each service restarting and downloading the plugin, _do not_ add the plugin using Halyard. Instead, configure the plugin in the service's local file. For example, if your plugin extends Orca, add configuration to your `orca-local.yml` file.
```
dorbin, line 207: In the above section, you have the download coming before the restart. If that's how it happens, maybe keep that same order here.

guides/user/plugins/index.md
aimeeu: I admit I ripped off that sentence from one of the engineers. I'll check.

guides/user/plugins/index.md
aimeeu: Bullet lists with periods look so wrong to me. I must be defaulting to what I learned in grade school English class in the 1970s (still remember diagramming sentences too).  However, since both the _Chicago Manual of Style_ and the _Google developer documentation style guide_ call for periods if the bullet point is a complete sentence, I can't really argue, can I?

dorbin: Yeah, that's right. 😁 
Numbered lists too.
I have a strong feeling toward getting rid of periods at the ends of list items like this. The period's just a delimited, and there's already the line break and bullet to delimit the phrases or sentences. BUT keeping the period when these are complete sentences is the accepted style out there. One advantage of doing it this way is that if we want to add a sentence to an item in a list like this, we can do so without now making the list less parallel. (Though I like to add those as list paragraphs, between the bullet or number items)

guides/user/plugins/index.md
aimeeu: Thanks for catching that! Download happens after restart.

guides/user/plugins/index.md
aimeeu: So... the engineers said that service restart times in a large installation are loooooong. Clouddriver can take an hour or more if it has lots of accounts configured. They also said that the community is actively working to shorten service restart times.

dorbin (approved):
:dango:

dorbin:
Shall I merge it, @aimeeu ?

317brian:
@dorbin Aimee is on PTO for a couple of weeks.

PR #1998, opened 2020-08-10
Updating the AWS EC2

anooptejt:
Adding detailed and correct steps to Configure AWS EC2 with Spinnaker

PR #2020, opened 2020-09-02
Move cfn template into repo

clareliguori:
Related to https://github.com/spinnaker/spinnaker.github.io/pull/1998, moving the CloudFormation template for the EC2 instructions back into the GitHub repo, so that the community can make fixes.  I've made a couple of fixes as well: 1) use a more secure option for storing and retrieving IAM user credentials, and 2) fix a broken output when an EC2 instance profile is not created.


dorbin (approved):
LGTM. Thanks Clare!

PR #2021, opened 2020-09-09
docs(community): Add Spinnaker Summit banner linking to CFP

dnilasor:
Replacing the outdated Gardening Days banner text from July with a new one that links to the CFP for Spinnaker Summit, open for another week. No need to add a new image here, just swapped the reference to an existing summit logo image from Greg Comstock.


317brian (approved):
LGTM

PR #2030, opened 2020-09-16
feat(docs): tidying up of artifacts docs

maggieneterval:
* feat(docs): re-order Triggering Pipelines sections 

  The new artifacts UI is now the default. Re-arrange navigation sections so Triggering Pipelines with new UI comes first, and change language for old UI from "Standard" to "Legacy" for consistency with elsewhere in the documentation.

* feat(docs): redirect from pipeline triggers documentation that refers to the legacy artifacts UI 

* feat(docs): change "scheduled for removal" to "removed" because the past is in the past

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -346,7 +346,24 @@ guides:
             url: /guides/user/pipeline/
           - title: Managing Pipelines
             url: /guides/user/pipeline/managing-pipelines/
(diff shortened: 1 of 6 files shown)


ezimanyi (approved):
🧹 🏺 📜 ❄️ 💯

dorbin (approved):
✨ 🌟 ⚓ 🗿 🍡

PR #2033, opened 2020-09-18
Update configure-redis-usage.md

snalvi:
300s = 5min

### DIFF
--- a/setup/productionize/caching/configure-redis-usage.md
+++ b/setup/productionize/caching/configure-redis-usage.md
@@ -23,7 +23,7 @@ To adjust how this caching happens, Clouddriver exposes a few properties that
 can be set in `~/.hal/$DEPLOYMENT/profiles/clouddriver-local.yml`:
 
 ```yaml
-# How many seconds (default 30) between runs of agent. Lowering this number
+# How many seconds (default 30s) between runs of agent. Lowering this number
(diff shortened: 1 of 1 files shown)


plumpy:
Thanks for the PR! Given the conflicting statements, I was curious if the default was actually 300 seconds or 600 seconds. I went to check and the answer is (...drumroll... :drum: ) [30 seconds](https://github.com/spinnaker/clouddriver/blob/e9ca4af27de478278edf36e925f735d5946d31a3/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/RedisConfigurationProperties.groovy#L29)!

So it turns on the docs are wrong twice. Do you think you could update it to say 30 seconds? Thanks so much!

snalvi (commented):
Interval is 30s by default
Timeout is 300s / 5m by default
https://github.com/spinnaker/clouddriver/blob/e9ca4af27de478278edf36e925f735d5946d31a3/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/RedisConfigurationProperties.groovy#L29

plumpy (approved):
Oh sorry, I was reading the diff wrong! I thought the link you changed was about the interval.

Thanks for fixing this!

snalvi:
Waiting on the merge to close off the PR

PR #2045, opened 2020-10-02
docs(community): Update conference ad banner

nikema-armory:
I replaced the copy in the `conference_ad_row` to remove the reference to the CFP and add the link to registration and the hackathon. We also changed the call to action from "Submit your proposal now" to "Register today."

The change was needed because the CFP is now closed.

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,8 +1,8 @@
 ---
 layout: home
 conference_ad_row:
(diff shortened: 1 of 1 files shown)


dnilasor (approved):
I tested this locally and it LGTM. Thanks @nikema-armory  for updating this and kicking off our promo for the Summit!

PR #2050, opened 2020-10-05
feat(build): Setup for autogenerated service javadocs 

robzienert:


reference/api/service-plugin-apis/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: single
+title:  "Service Plugin APIs"
+sidebar:
+  nav: reference
+---
+
+- Orca
+  - [master](/reference/api/service-plugin-apis/orca-api/master/orca-api/)
```
jonsie, line 9: May as well flush this out with Gate, Keel, and Echo?

Also, we now have the concept of platform level APIs in `kork-api`.  Worth adding a separate category for that?

robzienert: Not sure what you mean by "separate category"?

robzienert: I'll add the other services as their docs become available - I don't want to link to pages that don't exist.

PR #2058, opened 2020-10-09
feat(plugins): Core developer guide to plugins

robzienert:
This is a copy/paste of our internal plugins guide for core contributors. It currently is linking to our own getting started doc for plugins, but that overlaps with existing docs on spinnaker.io - I'd be happy to PR that as well if people are interested.

I'm headed to vacation, so I figured the raw copy/paste would suffice for now and I can come back to some comments and clean it up.

The spinnaker.io website doesn't do as nice of a job rendering the docs as Notion does, so things look kind of weird... but I don't think there's much to do about that without some CSS changes.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -513,6 +513,13 @@ guides:
             url: /guides/developer/plugin-creators/compatibility-testing/
           - title: Plugin Project Configuration
             url: /guides/developer/plugin-creators/project-config/
+      - title: Plugin Core Contributor Guide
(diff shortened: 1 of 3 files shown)


markjacksonfishing:
This is really nice!

claymccoy (approved):
Looks great!

robzienert:
Will merge now, can improve as we go (like updating docs to reflect new reality when Dan's kork changes land!)

PR #2061, opened 2020-10-16
Update x509 instruction with roles definition section

kirillbilchenko:
Update x509 client cert instructions with proper roles section


setup/security/authentication/x509/index.md
```diff
@@ -103,7 +102,7 @@ Encoding with any other OID can be done by editing the `openssl.conf`.
     keyUsage = nonRepudiation, digitalSignature, keyEncipherment
     1.2.840.10070.8.1 = ASN1:UTF8String:spinnaker-example0\nspinnaker-example1
     ```
-    The same rule as in step above is applied for role defention in this [step](#roles-defenition)
+    The same rule as in step above is applied for role defention in the step above
```
dorbin: s/defention/definition/

Also, which steps are you referring to? The way this reads now, the step above has the same rule as the step above.

kirillbilchenko: I was trying to add anchor for explanation about `The final line in this file `1.2.840.10070.8.1= ASN1:UTF8String:spinnaker-example0\nspinnaker-example1` is what matters for creating a client certificate with user role information, as anything after `UTF8String:` is encoded inside of the x509 certificate under the given OID.`

kirillbilchenko: @dorbin rephrased the explanation, I hope should be more clear

kirillbilchenko:
@dorbin  is it possible to merge this? or I can close pr?

dorbin:
Aughhh! Sorry, Kirill, I've been out of touch for a while, working on other stuff. 
These changes are fine, as far as I'm concerned, but I don't have time right now to test it. I just need to find someone to do a quick tech review of it. 
I'll try... @ezimanyi . Eric, can you take a quick look? Thanks!

ezimanyi:
I'm not overly familiar with generating SSL certificates, so I might not be the best person to review this.  But I'm not sure I understand why we need to re-define those fields in a separate extensions file when the are already defined in the above `openssl.conf`.

kirillbilchenko:
> I'm not overly familiar with generating SSL certificates, so I might not be the best person to review this. But I'm not sure I understand why we need to re-define those fields in a separate extensions file when the are already defined in the above `openssl.conf`.
Because if you are signing the server without extensions role-based auth is not been applied and the process is not working.

kirillbilchenko:
@ezimanyi @dorbin  I found the same info in the spinnaker slack https://spinnakerteam.slack.com/archives/C99CEV2MA/p1551126207007500

setup/security/authentication/x509/index.md
```diff
@@ -90,14 +90,23 @@ Encoding with any other OID can be done by editing the `openssl.conf`.
...
     >**Note:** If providing multiple groups, as in this example, separate them with a new line (`\n`). The new line `\n` shows as a `%0A` in the certificate.
 
-1. Generate a CSR for a new x509 certificate and the given `openssl.conf`:  
+1. Generate a CSR for a new x509 certificate and the given `openssl.conf`:
     ```
     openssl req -nodes -newkey rsa:2048 -keyout key.out -out client.csr \
         -subj "/C=US/ST=CA/L=Oakland/O=Spinnaker/CN=example@example.com" -config openssl.conf
     ```
+
+1. Create extention config file `extention.conf` to apply roles when signing the server requests.
```
ezimanyi:
```suggestion
1. Create extention config file `extension.conf` to apply roles when signing the server requests.
```

ezimanyi (commented):
I guess that seems reasonable, thanks for the explanation and link!  But I honestly am not sure I'm really familiar enough with this to sign off myself.

@dorbin maybe the @spinnaker/sig-security can take a quick look if you want a technical sign off?  It certainly seems reasonable but I don't know enough about the details of these requests to know for sure that this is the best workaround for the linked issue.

kirillbilchenko:
@ezimanyi @dorbin anyone on this for review?

ezimanyi:
Maybe @jasonmcintosh can provide a review from the Security SIG perspective.  This isn't something I'm familiar enough with to provide a review on myself.

jasonmcintosh:
Yeah I’ll take a look after Christmas break. I’m pretty familiar with x509 and OpenSSL. Can we add the sig security label?  Thanks!

setup/security/authentication/x509/index.md
```diff
@@ -90,14 +90,23 @@ Encoding with any other OID can be done by editing the `openssl.conf`.
...
+    [ v3_req ]
+    keyUsage = nonRepudiation, digitalSignature, keyEncipherment
+    1.2.840.10070.8.1 = ASN1:UTF8String:spinnaker-example0\nspinnaker-example1
+    ```
+    The same rule for the roles definition applied to this section, as it's explained in the first step of this section.
+
 1. Use the CA to sign the server's request. (If using an external CA, they do this for you.)
     ```
-    openssl x509 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out client.crt
+    openssl x509 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out client.crt -extensions v3_req  -extfile ./extension.conf
```
jasonmcintosh, line 109: This is actually required - just having it in the base config apparently doesn't work.  Whether it's required as a separate file or with-in the main file and you reference that tends to be a matter of preference.  It's mostly needing the config block in the file referenced via extfiles.

jasonmcintosh (approved):
Yeah this looks good.  The ONLY thing it might be useful to reference is documentation on openssl in general vs. having it locally.  And some openssl commands to validate the certs afterwards to show the OIDs.  BUT that should be another PR.

setup/security/authentication/x509/index.md
kirillbilchenko: for me in this case it's more clear what will be used for this specific command and can be easily checked, but you are right they are duplicating the content with the base file.

dorbin (approved):

PR #2063, opened 2020-10-19
docs(community): update gardening docs for Summit

dnilasor:
I'd like to propose this update to the Gardening area. I would like to remove some of the pages and link/defer to the repository documentation so we are not maintaining the doc in two places. I have also updated dates, graphics, and other basic info for the Summit event.


317brian (approved):
LGTM

PR #2064, opened 2020-10-19
docs(community): Remove outdated schedule from gardening/schedule 

prophen:
I removed the sched.com embed from the last gardening days event because that information is now outdated. This change is made in response to a suggestion by @brian-armory


317brian (approved):
Thanks, lgtm!

PR #2080, opened 2020-10-27
docs(GitRepo): tokenFile and username-password-file not supported

dperezro:
GitRepo doesn't support tokenFile and username-password-file, as seen in https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccount.java#L54

This java class probably is a copy of GitHubArtifactAccount.java which does support those.


317brian:
Hi @dperezro, thanks for opening this PR. The Halyard commands page is actually autogenerated from the code for https://github.com/spinnaker/halyard

E.g. 
https://github.com/spinnaker/halyard/blob/f71a745a4d5068efab147df3a4a3d4b37eabaa4a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/gitrepo/GitRepoAddArtifactAccountCommand.java#L42

dperezroCT:
Oh, so the PR should go there :) thanks!

dperezro:
Oh, so the PR should go there :) thanks!

PR #2083, opened 2020-10-29
docs(plugins): adding plugin framework v2 config updates

link108:


guides/user/plugins/index.md
```diff
@@ -230,6 +230,17 @@ The plugin developer should provide configuration details in YAML format. If not
...
+As of Spinnaker 1.23.0, listing extensions has been deprecated and configuration has been simplified. Plugin extension configurations have been moved and are now nested under the plugin itself, eg:
+```
+spinnaker:
+  extensibility:
+    plugins:
+      myPlugin:
+        config:
+          defaultMaxWaitTime: 7
+```
+
```
aimeeu: The spinnaker.extensibility.plugins.\<pluginId\>.extensions section is no longer needed, so plugin configuration in 1.23 is:

```
spinnaker:
  extensibility:
    plugins:
      <unique-plugin-id>:
        id: <unique-plugin-id>
        enabled: <true-false>
        version: <version>
        config: {}
```

Correct??  If yes, then perhaps replace the your specific example with the generic one.

Also, the "Plugin V2 Configuration Changes" is only for the "Plugin Configuration without Halyard" section, right? Or do we need to point out the 1.23 configuration change in the paragraph starting "Halyard does not support configuring plugins." right above the "Plugin configuration without Halyard" section.

aimeeu (changes requested):
A few questions

aimeeu (approved):

PR #2088, opened 2020-11-04
docs(halyard): add Java requirement for halyard

ronaldpetty:
Adding requirement (unclear which version though).

Installing halyard fails if you do not have Java installed which is more common with modern VMs (etc. Ubuntu 20.04 - AWS).

### DIFF
--- a/setup/install/halyard.md
+++ b/setup/install/halyard.md
@@ -42,7 +42,7 @@ Halyard runs on...
 
 Prerequisite:
 
-* Halyard requires Java to be installed
+* Halyard requires Java 11 to be installed
 
 1. Get the latest version of Halyard:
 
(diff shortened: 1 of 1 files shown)


setup/install/halyard.md
```diff
@@ -40,6 +40,10 @@ Halyard runs on...
 * Debian 8 or 9
 * macOS (tested on 10.13 High Sierra only)
 
+Prerequisite:
+
+* Halyard requires Java to be installed
```
317brian:
```suggestion
* Halyard requires Java 11 to be installed

317brian (approved):
LGTM, thanks for adding this requirement! Pretty annoying for a user to find out by hitting the error.

PR #2124, opened 2021-01-18
Remove Spinnaker Summit 2020

Nirmalyasen:
It is in the past and not correct for the landing page


aimeeu (approved):
Thanks @Nirmalyasen!!

PR #2126, opened 2021-01-22
spin-cli execute pipeline with required artifacts

Isakgicu:
https://github.com/spinnaker/spinnaker/issues/5727#issuecomment-765573984


317brian:
Thanks for the PR! @Chinikins can you give this a look since you're the spin CLI expert?

317brian:
ping @Chinikins

PR #2127, opened 2021-01-26
docs(spel): Add information on create data within pipelines using Parameters and Evaluate Variables stage in Pipeline Expression Guide

marqaaron-armory:
These commits not only add key information about the use of Parameter and the Evaluate Variables stage but also include the list of reserved names that cannot be used for either a parameter or an evaluated variable.  Also included is a standard warning and link to the list located in the Pipeline Expression Guide anywhere Parameters are referenced in other docs.


317brian (approved):
Thanks!

PR #2128, opened 2021-02-03
Update stageByRefId description

marcodoordash:
The description of the stageByRefId helper function has an example that uses #stage() instead of #stageByRefId()


aimeeu (approved):

aimeeu:
Thanks @marcodoordash!!

Issue #2153, opened 2021-03-17
https://spinnaker.io/community/stats/ needs some love


zomglings:
Page: [/community/stats/](https://www.spinnaker.io/community/stats/)

Feedback:
1. The link to the Linux Foundation Telemetry Data Collection and Usage Policy is broken: https://www.linuxfoundation.org/telemetry-data-collection-and-usage-policy/
2. The public dashboard simply shows a dataset configuration error:
![image](https://user-images.githubusercontent.com/8016073/111396421-4795d300-867c-11eb-8acd-968550c9713a.png)

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

conradkleinespel:
Who can we ask for information about how this is setup?

conradkleinespel:
@spinnakerbot remove-label to-be-closed

conradkleinespel:
Who can we ask for information about how this is setup?

PR #2186, opened 2021-06-22
docs(stats): Fix halyard command to enable/disable telemetry

conradkleinespel:
It seems that `hal config telemetry` has been renamed to `hal config stats`.

`hal config telemetry` does not work, as reported here: https://github.com/spinnaker/spinnaker/issues/6419

What's also a bit odd is that the stats page looks broken on my end: https://spinnaker.io/community/stats/

image


317brian (approved):
Thanks for catching this! I'm not quite sure who to ping for fixing the stats themselves though.

conradkleinespel:
I guess regarding the stats, it is already tracked here: https://github.com/spinnaker/spinnaker.github.io/issues/2153
Thanks to Johannes on Slack for pointing me to this issue.

